PR #47, opened 2021-11-26
Old password is not required for signed in users to change password

elitan:
hasura-auth-js PR: https://github.com/nhost/hasura-auth-js/pull/7

### DIFF
--- a/docker/dev/.env.example
+++ b/docker/dev/.env.example
@@ -3,7 +3,7 @@ POSTGRES_PASSWORD=secretpgpassword
 
 # HASURA
 HASURA_GRAPHQL_DATABASE_URL=postgres://postgres:secretpgpassword@postgres:5432/postgres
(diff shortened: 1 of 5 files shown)


nunopato (commented):
Are there any cases where it would make sense to provide the old password? if yes, then we should consider making it optional? If not, this is good to go.

elitan:
Yea maybe making the `oldPassword` optional is better.

My reasoning is, that I foresee us refactoring this password-change feature in the near future to support:

1. **`oldPassword` -** for people who know their old password
2. **Some ticket -** that comes from the fact that the user clicks on the password reset link in the email.

Basically, the user can change password if they either know their old password or recently clicked on a password reset link. Not, just by the fact that they are signed in.

However, having the `oldPassword` optional feels a bit... strange.

nunopato:
That sounds reasonable :ship:

nunopato (approved):

PR #51, opened 2021-12-02
Add `options` to reset password 

alveshelio:
I've also noticed a discrepancy between nhost-js-sdk type `ResetPasswordParams` and nhost-auth `userPasswordResetSchema` type. 
In nhost-js-sdk `ResetPasswordParams` is declared as follow:
```ts
export declare type ResetPasswordParams = {
    email: string;
    options?: {
        redirectTo?: string;
    };
};
```

However, in nhost-auth `userPasswordResetSchema` is defined as follow
```ts
export const userPasswordResetSchema = Joi.object({
  email: emailRule.required(),
});
```
lacking the object `options`. Also looking at `userPasswordResetHandler`, `BodyType` is defined as follow:
```ts
type BodyType = {
  email: string;
  options?: {
    redirectTo?: string;
  };
};
```
Indicating that it does accepts `options`.

This PR adds `options` to `userPasswordResetSchema`


nunopato (approved):
Thank you for opening the PR @alveshelio.

//c @elitan

elitan:
Thanks, @alveshelio! 

Before I dig deeper into this PR, you mention `nhost-js-sdk` in your description, however, `nhost-js-sdk` is related to Hasura Backend Plus only and not Hasura Auth.

All JavaScript SDK code related to Hasura Auth is in the following repositories:

- https://github.com/nhost/hasura-auth-js
- https://github.com/nhost/nhost-js

Are there any adjustments you want to make to this PR's code or descriptions?

elitan:
Ok, now looked through the code and it didn't match the description soo much :D 

Maybe my previous comment helped out with potential confusion.

Regarding this PR, it looks good. However, I think we should keep the current test and add your tests as additional tests. This way we test both with and without `options`.

elitan (changes requested):
Ok, now looked through the code and it didn't match the description soo much :D

Maybe my previous comment helped out with potential confusion.

Regarding this PR, it looks good. However, I think we should keep the current test and add your tests as additional tests. This way we test both with and without options.

alveshelio:
> Thanks, @alveshelio!
> 
> Before I dig deeper into this PR, you mention `nhost-js-sdk` in your description, however, `nhost-js-sdk` is related to Hasura Backend Plus only and not Hasura Auth.
> 
> All JavaScript SDK code related to Hasura Auth is in the following repositories:
> 
> * https://github.com/nhost/hasura-auth-js
> * https://github.com/nhost/nhost-js
> 
> Are there any adjustments you want to make to this PR's code or descriptions?

Maybe it's my lack of understanding of how all the ecosystem works. I though that `nhost-js`sdk` was a layer to help work with `hasura-auth`. 

I will create a new test for sending options and keep the other intact.

Thank you for the comments

alveshelio:
> Ok, now looked through the code and it didn't match the description soo much :D
> 
> Maybe my previous comment helped out with potential confusion.
> 
> Regarding this PR, it looks good. However, I think we should keep the current test and add your tests as additional tests. This way we test both with and without options.

I went ahead and refactored a bit the tests to avoid duplication and created a new test to pass `redirectTo` when changing password with a ticket

ctrlemo:
Looking forward to this fix.
Having the flexibility to be able to redirect users to url's different from the client url is important.

I also noticed that this endpoint also suffers from the same issue when I tested locally.
https://github.com/nhost/hasura-auth/blob/main/src/routes/user/email/send-verification-email.ts

**Error validating request body. "options.redirectTo" is not allowed.**

There might be more endpoints as well that need the same type of fix.

PR #52, opened 2021-12-02
fix(user): fix user schemas

alveshelio:
There was a typo in the `redirectTo` key in the `options` object for schemas `userEmailChangeSchema` and `userEmailSendVerificationEmailSchema`

Looking at type `BodyType` in `userEmailSendVerificationEmailHandler` and `userEmailChange` they both have `redirectTo` key in `options` object, not `redirecTo`.


test/routes/user/email.test.ts
```diff
@@ -109,4 +109,79 @@ describe('user email', () => {
...
+      .send({ email, password })
+      .expect(200);
+
+    expect(body.session).toBeTruthy();
+
+    if (!body.session) {
+      throw new Error('session is not set');
+    }
+
+    accessToken = body.session.accessToken;
```
nunopato, line 143: @alveshelio is there a reason to declare `accessToken` in line #119 if you are only using it here? 

`const accessToken = body.session.accessToken;`

test/routes/user/email.test.ts
```diff
@@ -109,4 +109,79 @@ describe('user email', () => {
...
+
+    // request to reset (to-change) email
+
+    const newEmail = 'newemail@example.com';
+
+    await request
+      .post('/user/email/change')
+      .send({ newEmail })
+      .expect(401);
```
elitan, line 152: Very nice that we're testing this! 👍

elitan (approved):
Looks really good!

test/routes/user/email.test.ts
alveshelio: No particular reason :P, sorry about that, I should have spent a little more time understanding how you guys are testing and specially how everything works :)
I've simply copied a test and adjusted for my use case.

alveshelio: I went ahead and refactored a bit the tests and moved some logic to avoid duplication and actually I've added tests for email verification. Let me know if this is good.

PR #67, opened 2022-01-18
clean code and streamline dev/test/build pipelines

plmercereau:
- check environment variables before anything else
- streamline tsconfig and path alias
- bump dependencies
- remove unused code e.g. prod-paths, inexistent path aliases, old methods...
- improve typings (pass lint script)
- remove unused dependencies e.g. esbuild, markdown-it
- develop locally rather than in a docker wrapper
- simplify package.json scripts
- use .env variables when possible e.g. cogegen, apollo
- add a basic developer's guide
- remove unfinished debugging-related console.log


docker-compose.yaml
```diff
@@ -0,0 +1,39 @@
...
+    image: nhost/postgres:12-v0.0.6
+    restart: always
+    volumes:
+      - ./docker/db_data/db:/var/lib/postgresql/data
+    environment:
+      POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-secretpgpassword}
+    ports:
+      - '5432:5432'
+  graphql-engine:
+    image: hasura/graphql-engine:v2.1.0
```
elitan, line 13: Let's use `v2.1.1` which is the latest and the one we use for customers in production.

elitan (changes requested):
Nice clean-up!

docker-compose.yaml
plmercereau: ok

PR #72, opened 2022-01-19
locales templates and passwordless

plmercereau:
Streamline the use of email templates:
- some routes were not picking the locale in this order: option -> user locale -> application locale
- fallback to the default application locale when the requested locale doesn't exist

...and don't send passwordless login email when the user is deactivated


src/routes/signin/passwordless/email.ts
```diff
@@ -127,7 +133,7 @@ export const signInPasswordlessEmailHandler = async (
       email,
       ticket,
       redirectTo,
-      locale: user.locale,
+      locale: options?.locale ?? user.locale ?? ENV.AUTH_LOCALE_DEFAULT,
```
elitan, line 136: I'm not 100% sure we should change this.

Is there a use case to send an email with another locale than the user's locale?

Also, the `user.locale` should always be set so the last `?? ENV.AUTH_LOCALE_DEFAULT` is not needed?

Another argument for keeping just the `locale: user.locale` and not add extra options are generally fewer things to worry about for us. Fewer docs, fewer tests, etc.

Yes, it's also less flexible for the end-user not to include this extra `locale` option, but until there is a big demand to change it I think we should keep it as it is.

elitan (commented):
My comment is more generally on flexibility vs product surface area. My argument is that let's try to keep things less flexible until we really need to, so we can save time on things like do docs, tests, support and more.

src/routes/signin/passwordless/email.ts
plmercereau: a user could have been registered in `fr`, but swapped in `nl` in its UI. The frontend decides which locale to use (default going to the user locale if none is given).
Yes, at this time of the code, the user locate is supposed to be always set

nunopato: `options.locale` is just a simple way of overriding the user's locale in the context of a given request. I can't think of a use-case for that but I don't see a problem with it either.

elitan: > a user could have been registered in fr, but swapped in nl in its UI. The frontend decides which locale to use (default going to the user locale if none is given).

We can always come up with imagined scenarios, but how often will that happen? If the user switches locale in the UI the user's locale should be updated in the database. Problem solved.

> options.locale is just a simple way of overriding the user's locale in the context of a given request. 

Correct

> I can't think of a use-case for that but I don't see a problem with it either.

The "problem" I see is more product surface area leading to more docs, more support, more to test, more potential bugs, more code.

My point is: I don't think we should add complexity to the product if we don't have to because there is always a cost to it. Unless we really need it, let's not have it.

plmercereau: addressed in last commit: locale is available as an option only when the user sends his credentials for the first time e.g. email+password signup

PR #77, opened 2022-01-24
fix(metadata): show column values when the column name is the same as the graphql field name

### LINKED ISSUE #76: Hasura Console data browser issue with id metadata
We set the `id` column to `id` using Hasura metadata [here](https://github.com/nhost/hasura-auth/blob/main/src/metadata.ts#L212). This has no real impact. The field name will be `id` with and without the metadata.

There is a [bug](https://github.com/hasura/graphql-engine/issues/8020) in the Haura Console with setting the `id` metadata, effectively making the Hausra Console unable to use with data in these tables.

So I'm thinking we can remove the `id` metadata for now.

Related issue:

https://github.com/nhost/nhost/issues/121

plmercereau:
Workaround of a bug in Hasura: https://github.com/hasura/graphql-engine/issues/8020

fix #76

### DIFF
--- a/src/metadata.ts
+++ b/src/metadata.ts
@@ -209,15 +209,18 @@ export const applyMetadata = async (): Promise<void> => {
         delete_by_pk: 'deleteUser',
       },
       custom_column_names: {
-        id: 'id',
+        // * See: https://github.com/hasura/graphql-engine/issues/8020
+        // id: 'id',
         created_at: 'createdAt',
         updated_at: 'updatedAt',
         last_seen: 'lastSeen',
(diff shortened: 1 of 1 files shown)


elitan:
@plmercereau Did you test this PR with Hasura Console 2.1.x? if not, can you please do it to make sure we're fixing the issue?

I'm pretty sure we do, but I just want to confirm.

elitan (approved):
Changes looks good

plmercereau:
Sure, I did test on Hasura Concole 2.1.1 :)

PR #78, opened 2022-01-24
JWT custom claims and JSONB `auth.users.custom` column

plmercereau:
1. Custom column
A basic JSONB column in the `auth.users` table, that is passed on as an option on registration:
```json
{
  "email": "bob@bob.com",
  "passord": "12345678",
  "options": {
    "custom": {
      "first_name": "Bob"
    }
  }
} 
```

2. Custom JWT claims
Introduction of a new `AUTH_JWT_CUSTOM_CLAIMS` environment variable, for instance
```
AUTH_JWT_CUSTOM_CLAIMS={"organisation-id":"profile.organisation.id", "project-ids":"profile.contributesTo.project.id"}
````

Will generate and fetch the following GraphQL query:
```graphql
{
  user(id:"<user-id>") {
    profile {
      organisation {
        id
      }
      contributesTo {
        project {
          id
        }
      }
    }
  }
}
```
It will then use the same expressions e.g. `profile.contributesTo.project.id` to evaluate the result with [JSONata](https://jsonata.org/), and possibly transform arrays into Hasura-readable, PostgreSQL arrays.
Finally, it adds the custom claims to the JWT in the `https://hasura.io/jwt/claims` namespace:

```json
{
  "https://hasura.io/jwt/claims": {
    "x-hasura-organisation-id": "8bdc4f57-7d64-4146-a663-6bcb05ea2ac1",
    "x-hasura-project-ids": "{\"3af1b33f-fd0f-425e-92e2-0db09c8b2e29\",\"979cb94c-d873-4d5b-8ee0-74527428f58f\"}",
    "x-hasura-allowed-roles": [ "me", "user" ],
    "x-hasura-default-role": "user",
    "x-hasura-user-id": "121bbea4-908e-4540-ac5d-52c7f6f93bec",
    "x-hasura-user-isAnonymous": "false"
  }
  "sub": "f8776768-4bbd-46f8-bae1-3c40da4a89ff",
  "iss": "hasura-auth",
  "iat": 1643040189,
  "exp": 1643041089
}
```


src/utils/jwt/custom-claims.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  return Object.entries(ENV.AUTH_JWT_CUSTOM_CLAIMS).reduce<
+    Record<string, unknown>
+  >((aggr, [name, path]) => {
+    const expression = jsonata(path);
+    aggr[`x-hasura-${name}`] = escapeValueToPg(
+      expression.evaluate(user, expression)
+    );
+    return aggr;
+  }, {});
```
elitan, line 75: This code could use some comments or increase readability for the future.

migrations/00002_custom_user_fields.sql
```diff
@@ -0,0 +1,5 @@
+-- start a transaction
+BEGIN;
+ALTER TABLE auth.users
+ADD COLUMN custom JSONB;
```
elitan, line 4: Any specific reason to call this `custom` instead of `metadata`?

elitan (commented):
Huge PR! 🚀 

One question: What happens if the `AUTH_JWT_CUSTOM_CLAIMS` variable is incorrectly set. E.g. a relationship does not exist on the `user` that is used in the env var.

I'm also curious to hear some feedback from @nunopato.

plmercereau:
> One question: What happens if the `AUTH_JWT_CUSTOM_CLAIMS` variable is incorrectly set? For E.g. a relationship does not exist on the `user` that is used in the env var.

I adjusted the code and added some tests to make sure the custom claims are ignored if the configuration is somehow incorrect, but to log server warnings (on startup if the env var is malformed, and then on runtime)

src/utils/jwt/custom-claims.ts
plmercereau: Fair point 😬 
I just pushed [a commit to explain the algorithm in detail](https://github.com/nhost/hasura-auth/pull/78/commits/69ca45d70645a9d3a915cebe52ea01f1dcbdf7f9#diff-3fb2c00b8b08f1ae2026d67d8dac7ba2c23e3b5478d4390f96165c17c0a6af49)

migrations/00002_custom_user_fields.sql
plmercereau: Naming things is always tricky, I'm fine with `metadata`, no strong opinion on the matter - `metadata` then?

elitan: Yea it's a good question.

I see this column more as a way for developers to attach data during signup rather than storing general metadata for users. In fact, I'm not sure we should encourage developers to use this unstructured JSON `metadata` column to store user metadata. User metadata should probably be stored in a `public.profiles` table.

Supabase has a `metadata` column on their `auth.users` table but they mention this in [their docs](https://supabase.com/docs/reference/javascript/auth-update):

> User metadata: It's generally better to store user data in a table inside your public schema (i.e. public.users). Use the update() method if you have data which rarely changes or is specific only to the logged in user.

The naming of this column could have an impact on the way developers use the column and the way we document the recommended usage of the column.

elitan: @nunopato Any thoughts on this?

nunopato: I think the `metadata` column makes sense for data that does not change often, consists of a single value, and is known at signup time (e.g., `family_name`, `age`, etc). Having a separate table makes sense for data like a user's address where you would have to store multiple values (e.g., street, address, etc). I think this is how we should document its usage.

nunopato (commented):
Great work @plmercereau! I think the implementation for custom claims is very smart. Would it make sense to extend it to also allow for the custom data solved with 1?

elitan:
> Great work @plmercereau! I think the implementation for custom claims is very smart. Would it make sense to extend it to also allow for the custom data solved with 1?

Do you mean also allow to add custom claims with values from `users.metadata` column? I think that feature can be postponed until we see a need for it.

migrations/00002_custom_user_fields.sql
plmercereau: Let's change to `metadata` then!

plmercereau:
> > Great work @plmercereau! I think the implementation for custom claims is very smart. Would it make sense to extend it to also allow for the custom data solved with 1?
> 
> Do you mean also allow to add custom claims with values from `users.metadata` column? I think that feature can be postponed until we see a need for it.

Yes, let's set is aside for now, also because we cannot really use the same algorithm:
- in custom claims, we original expression e.g. `profile.projects.id` expands the data graph query, to then transform and aggregate it. The algo goes 1. expression 2. graphql data 3. jsonata transformation
- in user metadata, it starts from a json object. The transformation and the graphql mutation may not be the same.

We'll come to that :)

elitan (approved):

philefstat:
@plmercereau is this live on cloud-hosted nhost?

noverby:
@plmercereau can this also be used to define roles inside an organization?

ryoid:
Hello, I am trying to have an allowed list of organization ID claims for each user, similar to "x-hasura-project-ids" with a PostgresSQL array, which I then use in my Hasura permissions 

`{"id":{"_in":"X-Hasura-Allowed-Org-Ids"}} `

However, I encountered an issue with the setup getting a data-exception error if the claim result was a single entry. Is there a recommended way to have this setup that would handle zero/one/many ids?

elitan:
> @plmercereau is this live on cloud-hosted nhost?

Yes, it is. Check February's update here: https://nhost.io/blog/nhost-february-2022

> @plmercereau can this also be used to define roles inside an organization?

No, the custom claim feature has nothing to do with roles. You can define roles for users by modifying the database directly (`auth.user_roles`).

elitan:
> Hello, I am trying to have an allowed list of organization ID claims for each user, similar to "x-hasura-project-ids" with a PostgresSQL array, which I then use in my Hasura permissions
> 
> `{"id":{"_in":"X-Hasura-Allowed-Org-Ids"}} `
> 
> However, I encountered an issue with the setup getting a data-exception error if the claim result was a single entry. Is there a recommended way to have this setup that would handle zero/one/many ids?

Oh, I don't think we support PostgreSQL arrays actually (but we should). Would you mind creating an issue for that?

The way you can solve this is to create a separate table instead of an array. Like:

```
Table name: allowed_orgs (don't know if this is a perfect name for you)

- id
- organization_id
- user_id
```

And have a relationship between `auth.users` and `public.allowed_orgs` so that you can get a user's allowed org id's with something like:

```
user(id) {
  id
  displayName
  allowedOrgs {
    organization_id
  }
}
```

That way you can set the custom claim like this: `x-hasura-allowed-org-ids: user.allowedOrgs.organization_id`.

danahartweg:
I'm also seeing the same issue as @ryoid , but when using the structure as outlined above by @elitan . In the case where there is only *one* record in the `allowedOrgs` array relationship, the `organization_id` is returned by itself, not inside an array.

PR #89, opened 2022-02-02
fix: allow redirect urls in Oauth that starts with the one defined in the server

plmercereau:
apply isValidRedirectTo in Oauth as well, and simplify its signature


elitan (approved):
🚢 IT

PR #101, opened 2022-03-02
Add swagger endpoint

plmercereau:


elitan (approved):

szilarddoro (approved):
Looks good 👍🏼

One thing I'd like to mention here is that probably we should get rid of commented code blocks like `// .required(); // .allowedDomains();`, etc. if there is no particular reason to keep them in our codebase.

plmercereau:
> One thing I'd like to mention here is that probably we should get rid of commented code blocks like `// .required(); // .allowedDomains();`, etc. if there is no particular reason to keep them in our codebase.

Yup, that's something I will work on next cycle

Issue #80, opened 2022-01-25
Handle locale, allowedRoles and defaultRole too. It currently handle only redirectTo


regalstreak:
Currently we're supporting only `redirectTo`, add support for `locale`, `allowedRoles` and `defaultRole`

https://github.com/nhost/hasura-auth/blob/main/src/routes/signin/providers/utils.ts#L255-L258

Related: nhost/nhost#188

elitan:
Fixes: https://github.com/nhost/nhost/issues/188

PR #108, opened 2022-03-14
fix: provider requests signup data and redirectTo

elitan:
This PR fixes:

- Add signup data to request (like metadata, roles, displayName etc)
- Use correct variable name for redirecting: `redirectTo` instead of `redirectUrl`

fixes: https://github.com/nhost/hasura-auth/issues/80, https://github.com/nhost/nhost/issues/188

### DIFF
--- a/migrations/00004_provider-request-generic.sql
+++ b/migrations/00004_provider-request-generic.sql
@@ -1,0 +1,13 @@
+-- start a transaction
+BEGIN;
+
+ALTER TABLE
+  auth.provider_requests
+DROP COLUMN redirect_url;
+
+ALTER TABLE
+  auth.provider_requests
+ADD
+  COLUMN metadata JSONB;
+
+COMMIT;
--- a/migrations/00004_provider-request-options.sql
+++ b/migrations/00004_provider-request-options.sql
@@ -8,6 +8,6 @@ DROP COLUMN redirect_url;
 ALTER TABLE
   auth.provider_requests
 ADD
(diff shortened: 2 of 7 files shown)


migrations/00004_provider-request-generic.sql
```diff
@@ -0,0 +1,13 @@
...
+BEGIN;
+
+ALTER TABLE
+  auth.provider_requests
+DROP COLUMN redirect_url;
+
+ALTER TABLE
+  auth.provider_requests
+ADD
+  COLUMN metadata JSONB;
```
elitan, line 11: Either we use a general `metadata JSONB` column or create individual columns for:

- redirectTo
- displayName
- locale
- defaultRole
- allowedRoles
- metadata

elitan: It's also safe to do this change later if we want to.

plmercereau: I would rather name the column options than metadata as it captures all registration options, including options.metadata, options.redirectTo etc

plmercereau (approved):

nunopato (commented):
We spoke about this offline, but leaving my comments here as well. 

Since `redirectTo` makes sense to be a `signIn` option, we should have it per provider request, hence be a column. 

The other options are `signUp` options, meaning they should only be set the first time the user uses a provider and only for context (the data is also stored in users). A JSON blob field seems the right choice.

nunopato (approved):
lgtm

PR #109, opened 2022-03-14
feat: error redirects

elitan:
Instead of showing the error message for the user like "Invalid or expired verification ticket" Hasura Auth will redirect the user to the client with error information as query parameters. Like this:

`https://example.com?error=invalidOrExpiredVerificationTicket&errorDescription=Invalid%20or%20expired%20verification%20ticket`

This way, the app can show a nice-looking and relevant error message for the user.

### DIFF
--- a/src/routes/verify/verify.ts
+++ b/src/routes/verify/verify.ts
@@ -1,3 +1,4 @@
+import { generateRedirectUrl } from '@/utils';
 import { gqlSdk } from '@/utils/gqlSDK';
 import { getNewRefreshToken } from '@/utils/tokens';
 import { Request, Response } from 'express';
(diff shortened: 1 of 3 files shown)


src/utils/index.ts
```diff
@@ -0,0 +1,35 @@
+const generateRedirectUrl = (
+  redirectTo: string,
+  queryParameters: { [key: string]: string },
+  hashTag?: string
+): string => {
+  let finalRedirectTo = redirectTo;
+
+  // add query paramters
+  // add first ? or & depending on of there are already query parameters
```
plmercereau, line 9: why are we using both `#` and `?`? Wouldn't it be easier to put everything after `#` next to `refreshToken` and `type`?

elitan: There is generally better support to get (and set) query parameters (`?` and `&`) than getting info from a hashtag (`#`).

The only reason we put the refresh token after a hashtag (instead of a regular query parameter) is because of security concerns. Data after a hashtag can not be read by the server, won't be saved in server logs, etc.

src/routes/verify/verify.ts
```diff
@@ -11,18 +12,29 @@ export const verifyHandler = async (
...
   const redirectTo = req.query.redirectTo as string;
 
+  if (!redirectTo) {
+    return res.boom.badRequest('Missing redirectTo');
+  }
+
   if (!ticket) {
-    return res.boom.badRequest('Missing ticket');
+    const redirectUrl = generateRedirectUrl(redirectTo, {
+      error: 'missingVerificationTicket',
```
nunopato, line 21: Given that this is considered an object or type, would it make sense to capitalize the first letter (`MissingVerificationTicket`) ?

nunopato (approved):

plmercereau (approved):

PR #110, opened 2022-03-14
test: make env variables available in tests

alveshelio:
This PR fixes the issue where env variables are not available in the tests.


plmercereau:
Merci!

PR #131, opened 2022-04-01
Insert the role in `auth.roles` when a new role is added to the Hasura metadata

### LINKED ISSUE #103: automatically add role to `auth.roles` when a new role is created in hasura metadata
See
https://github.com/nhost/nhost/discussions/31

plmercereau:
@nunopato we have to make sure the Postgresql user that Hasura Auth uses to connect is authorized to add an event trigger on the `hdb_catalog` schema.

See [this discussion](https://github.com/nhost/nhost/discussions/31)

Closes #103

### DIFF
--- a/migrations/00005_create-roles-from-metadata.sql
+++ b/migrations/00005_create-roles-from-metadata.sql
@@ -1,0 +1,22 @@
+CREATE OR REPLACE FUNCTION auth.update_roles_from_metadata () RETURNS TRIGGER LANGUAGE plpgsql AS $$ BEGIN
+INSERT INTO auth.roles(role)
+SELECT DISTINCT jsonb_array_elements_text(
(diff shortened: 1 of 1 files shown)


docker-compose.yaml
```diff
@@ -1,7 +1,7 @@
 version: '3.6'
 services:
   postgres:
-    image: postgres
+    image: postgres:12
```
plmercereau, line 4: I set the version to 12 as I struggled with a Postgres syntax error in the GH action (postgres v12) that was not happening when testing locally (latest Postgres version).
It turned out I was using `CREATE OR REPLACE TRIGGER`, which was introduced in Postgres v14

szilarddoro:
I'm closing this PR because it's stale.

PR #134, opened 2022-04-03
Revert 00002 migration name to previous name when migration fails because of it

### LINKED ISSUE #129: Hasura auth container fails to start after upgrade from v0.2 to v0.5
Migrations are applied correctly on a fresh db, seems to break on existing db

Possibly related to https://github.com/nhost/hasura-auth/commit/445ecb48793009399e5f5d006143c984090de594

```
> hasura-auth@0.5.0 start /app
> node ./dist/start.js

{"level":30,"time":1648739886630,"pid":18,"hostname":"hasura-auth-7fc5bc477c-nh25w","msg":"Hasura is ready"}
(node:18) UnhandledPromiseRejectionWarning: Error: Migration failed. Reason: Hashes don't match for migrations '00002_custom-user-fields.sql'.
This means that the scripts have changed since it was applied.
at /app/node_modules/.pnpm/@djgrant+postgres-migrations@4.1.0-rc.1/node_modules/@djgrant/postgres-migrations/dist/migrate.js:63:27
at processTicksAndRejections (internal/process/task_queues.js:95:5)
at async /app/node_modules/.pnpm/@djgrant+postgres-migrations@4.1.0-rc.1/node_modules/@djgrant/postgres-migrations/dist/with-lock.js:24:28
at async applyMigrations (/app/dist/migrations.js:16:9)
at async start (/app/dist/server.js:38:5)
```

plmercereau:
@nunopato @elitan we modified migration file names in previous versions
It turns out it was not a good idea: while Hasura migration system seems to ignore such name changes, `@djgrant/postgres-migrations` is not happy with it as it checks hashes from file names.

This fix is not elegant, but it's the only way that I've seen so far to handle the following cases:

- migration from 0.2 (`_` name) to further versions (`-` name)
- migration from 0.3 (`-` name) to further versions (`-` name)

closes #129


src/migrations.ts
```diff
@@ -16,6 +18,30 @@ export async function applyMigrations(): Promise<void> {
     await migrate({ client }, './migrations', {
       migrationTableName: 'auth.migrations',
     });
+  } catch (error: any) {
+    const userFieldsMigration = path.join(
```
elitan, line 22: Could you add an inline comment so we know why this code is here when we look at this 3+ months from now

elitan (changes requested):

elitan (approved):

mmmmillar:
Hi, is this likely to go in soon? I haven't been able to update because of it. Thanks

PR #140, opened 2022-04-14
Perf/startup

plmercereau:
- improve startup with async imports
- Improve logging on startup

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -1,58 +1,45 @@
-import express from 'express';
-import helmet from 'helmet';
-import { json } from 'body-parser';
-import cors from 'cors';
-import passport from 'passport';
-import { Server } from 'http';
+import { ENV } from '@/utils/env';
+import { logger, LOG_LEVEL } from './logger';
 
-import { ENV, waitForHasura } from '@/utils';
-import { applyMigrations } from './migrations';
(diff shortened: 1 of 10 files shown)


src/server.ts
```diff
@@ -1,58 +1,45 @@
...
+export const start = async () => {
+  logger.info(`Log level: ${LOG_LEVEL}`);
+
+  /**
+   * Async imports allow to dynamically import the required modules,
+   * and therefore only import modules that are required, when they are required.
+   * It decreases the loading time on startup.
+   * If we decide to apply migrations/metadata in a separate moment e.g. k8s initContainer,
+   * it will allow hasura-auth to load modules only required for migrations/metadata,
+   * or not to load these modules when auto migration/metadata is disabled.
```
elitan, line 13: This feels Nhost-specific and should probably not be in the Hasura Auth source code. Maybe we can store this information somewhere else?

elitan (changes requested):

src/server.ts
plmercereau: I rephrased the comment so it does not imply it is related to our infra, but in the way hasura-auth can be generally orchestrated

elitan (approved):

PR #142, opened 2022-04-19
fix: correct redirectTo

### LINKED ISSUE #137: [0.6.0] All social logins are broken, including discord
Example flow on logging in with Discord
- http://localhost:1337/v1/auth/signin/provider/discord?redirectTo=http%3A%2F%2Flocalhost%3A3000%2Fprofile, 302 redirect to
-  http://localhost:3000/profile?error=invalid-request&errorDescription=%22redirectTo%22%20must%20be%20one%20of%20[https://midnightsociety.com/claws-nft,%20https://vsm-nft-fe.vercel.app/claws-nft,%20http://localhost:3000/claws-nft]

hasura-auth needs E2E testing with a FE sample app. with say Playwright

plmercereau:
Allow redirection either starting with AUTH_CLIENT_URL or being strictly included in
AUTH_ACCESS_CONTROL_ALLOWED_REDIRECT_URLS. When no redirectTo is set, falls back to AUTH_CLIENT_URL

fix #137

### DIFF
--- a/src/utils/env.ts
+++ b/src/utils/env.ts
@@ -93,7 +93,7 @@ export const ENV = {
 
   // CLIENT
   get AUTH_CLIENT_URL() {
-    return castStringEnv('AUTH_CLIENT_URL', '');
+    return castStringEnv('AUTH_CLIENT_URL', '').toLocaleLowerCase();
   },
 
   // SIGN UP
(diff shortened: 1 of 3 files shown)


elitan:
Why do we strictly check `AUTH_ACCESS_CONTROL_ALLOWED_REDIRECT_URLS`?

I'm asking because I like the previous behavior better, to check if the `redirectUrl` starts with a value in `AUTH_ACCESS_CONTROL_ALLOWED_REDIRECT_URLS`.

plmercereau:
> Why do we strictly check `AUTH_ACCESS_CONTROL_ALLOWED_REDIRECT_URLS`?
> 
> I'm asking because I like the previous behavior better, to check if the `redirectUrl` starts with a value in `AUTH_ACCESS_CONTROL_ALLOWED_REDIRECT_URLS`.

👍
Added the suggestion to this PR.

I couldn't find where it behaved like this in the past, [v0.1.0](https://github.com/nhost/hasura-auth/blob/e991f00bd241f6fd33eef2e9efd79041bab800b8/src/helpers.ts#L150) and [mid-January versions](https://github.com/nhost/hasura-auth/blob/0b764255e02f0f0c3a72f19863f947403dbef56d/src/helpers.ts#L148) where already testing strict equality rather than values that start with any of the array items.

elitan:
Oh, sorry. You're right.

Then the question is if we should keep this behavior or not. 

I'm thinking it's better to check if it starts with the value, instead of checking strictly.

This means the same behavior for both `AUTH_CLIENT_URL` as for `AUTH_ACCESS_CONTROL_ALLOWED_REDIRECT_URLS`.

PR #145, opened 2022-04-27
Fix/allow unknown query parameters

plmercereau:
- Allow unknown query parameters
- Include `redirectTo`'s query parameters in the email link 
- Improve tests


nunopato (approved):
looks good :+1:

Issue #65, opened 2022-01-15
New refresh token expiration approach


elitan:
Every time a user signs in or refresh their access token (currently happens every 15 minutes) the existing refresh token gets deleted and a new refresh token is saved in the database and returned to the client.

This does not have any clear benefit. Instead it comes with some downsides:
- Race condition issues (the same refresh token is being used in two browser tabs at the same time to get a new access token)
- Extra operations on the database.

Instead, I propose that we change this strategy to how Firebase ([link](https://firebase.google.com/docs/auth/admin/manage-sessions)) handles refresh tokens:

- Refresh tokens never expire **except** if:
- The user gets deleted
- The user updates their email
- The user updates their password

kratam:
Do you need anything to move this forward?

dohomi:
this would be very helpful, currently the refresh expiration is way to short.

PR #156, opened 2022-05-06
feat: add JWT verify endpoint and getToken improvements

### LINKED ISSUE #83: Endpoint to verify a JWT is valid
cc @dbarrosop 
POST endpoint: `/token/verify`
in the body:
```json
{
  "token": "<the JSON web token>"
}
```
2xx http response if JWT is verified, 4xx response if not

L3K0V:
- Partially solves #65
- Closes #83

### DIFF
--- a/src/routes/token/token.ts
+++ b/src/routes/token/token.ts
@@ -14,20 +14,12 @@ export const tokenHandler: RequestHandler<
 > = async (req, res) => {
   const { refreshToken } = req.body;
 
-  // set expiresAt to now + 10 seconds.
-  // this means the refresh token is available for 10 more seconds to avoid race
-  // conditions with multiple request sent by the same client. Ex multiple tabs.
-  const expiresAt = new Date();
-  expiresAt.setSeconds(expiresAt.getSeconds() + 10);
-
-  // get user and set new expiresAt on the used refreshToken
   const refreshTokens = await gqlSdk
-    .getUsersByRefreshTokenAndUpdateRefreshTokenExpiresAt({
+    .getUsersByRefreshTokenOld({
       refreshToken,
-      expiresAt: expiresAt,
     })
     .then((gqlres) => {
-      return gqlres.updateAuthRefreshTokens?.returning;
+      return gqlres.authRefreshTokens;
     });
 
   if (!refreshTokens) {
@@ -39,6 +31,7 @@ export const tokenHandler: RequestHandler<
   }
 
   const user = refreshTokens[0].user;
+  const currentRefreshToken = refreshTokens[0].refreshToken;
 
   if (!user) {
     return sendError(res, 'invalid-refresh-token');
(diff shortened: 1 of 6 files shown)


src/routes/token/token.ts
```diff
@@ -39,6 +31,7 @@ export const tokenHandler: RequestHandler<
   }
 
   const user = refreshTokens[0].user;
+  const currentRefreshToken = refreshTokens[0].refreshToken;
```
elitan, line 34: I think this can simply be `const currentRefreshToken = refreshToken` from L15

src/routes/token/token.ts
```diff
@@ -14,20 +14,12 @@ export const tokenHandler: RequestHandler<
...
   const refreshTokens = await gqlSdk
-    .getUsersByRefreshTokenAndUpdateRefreshTokenExpiresAt({
+    .getUsersByRefreshTokenOld({
       refreshToken,
-      expiresAt: expiresAt,
     })
     .then((gqlres) => {
-      return gqlres.updateAuthRefreshTokens?.returning;
+      return gqlres.authRefreshTokens;
     });
```
elitan, line 24: I think we should also extend the expiration time of the refreshToken when its being used.

The default expiration time for a refresh token is 30 days. If the user keeps using the refresh token to get new access tokens the user will be signed out after 30 days because that's what the refresh token expires.

Ideally, the refresh token should reset its expiration time every time it's being used.

Does it make sense?

src/routes/token/verify.ts
```diff
@@ -0,0 +1,24 @@
...
+export const verifyTokenHandler: RequestHandler<
+  {},
+  {},
+  { token?: string }
+> = async (req, res) => {
+  const authorization = req.body.token || req.headers.authorization;
+
+  try {
+    getClaims(authorization);
+    return res.send(ReasonPhrases.OK);
```
elitan, line 20: I can't remember seeing this pattern elsewhere (`ResponsePhrases.OK`). Is this consistent with what we're doing elsewhere?

elitan (commented):
Thanks for this PR. I added some minor comments.

src/routes/token/verify.ts
L3K0V: Yes, of course! I personally don't like it, but can be improved/changed in a different PR

image

src/routes/token/token.ts
L3K0V: Totally make sense. I will adjust to be refreshed when used to obtain new token.

I left the default expiration time to 30 days because I think:
1. 30 days expiration time by default is reasonable
2. Is configurable, so everybody can adjust it whatever they feel comfortable with (eg. 1 year)

Updating the Refresh token expiration is the key here, thanks!

L3K0V: Addressed in https://github.com/nhost/hasura-auth/pull/156/commits/1c6fc32ff09af2991df067948e30878d5054e1fc

src/routes/token/verify.ts
elitan: Perfect! But, light mode?? 😅

src/routes/token/token.ts
elitan: Yea, the default expiration time should still be 30 days. Perfect.

src/routes/token/verify.ts
L3K0V: I knew someone will make a note of that! 🤣 It's synced with macOS system appearance

elitan (approved):

PR #167, opened 2022-05-24
feat: allow any `redirectTo` when no `AUTH_CLIENT_URL` is set

plmercereau:
Up to now, `AUTH_CLIENT_URL` had to be set to support transactional emails and Oauth. Now, when not
set, such feature will work as long as the `redirectTo` option is sent over by the client to Hasura
Auth. It is usefull when one wants any client to work.

Rationale:
- users don't need to determine what their client url would be before using "basic" features such as email verification or magic links
- users can use the same backend for both their `localhost` and production frontend
- run examples on stackblitz/codepen et al.

### DIFF
--- a/src/routes/signin/passwordless/email.ts
+++ b/src/routes/signin/passwordless/email.ts
@@ -115,7 +115,6 @@ export const signInPasswordlessEmailHandler: RequestHandler<
       redirectTo: encodeURIComponent(redirectTo),
       locale: user.locale ?? ENV.AUTH_LOCALE_DEFAULT,
       serverUrl: ENV.AUTH_SERVER_URL,
-      clientUrl: ENV.AUTH_CLIENT_URL,
     },
   });
 
--- a/src/routes/signup/email-password.ts
+++ b/src/routes/signup/email-password.ts
@@ -122,7 +122,6 @@ export const signUpEmailPasswordHandler: RequestHandler<
         redirectTo: encodeURIComponent(redirectTo),
         locale: user.locale,
         serverUrl: ENV.AUTH_SERVER_URL,
-        clientUrl: ENV.AUTH_CLIENT_URL,
       },
     });
   }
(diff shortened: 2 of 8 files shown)


src/routes/signin/passwordless/email.ts
```diff
@@ -115,7 +115,6 @@ export const signInPasswordlessEmailHandler: RequestHandler<
       redirectTo: encodeURIComponent(redirectTo),
       locale: user.locale ?? ENV.AUTH_LOCALE_DEFAULT,
       serverUrl: ENV.AUTH_SERVER_URL,
-      clientUrl: ENV.AUTH_CLIENT_URL,
```
elitan, line 118: I don't think we should delete this env var for emails because:

- It's a breaking change.
- It's useful to create generic links to the frontend app in the email.

elitan (changes requested):

src/routes/signin/passwordless/email.ts
plmercereau: sure, I reverted
this variable could however be inconsistent with redirectTo when it points to another client url

elitan: I think of `client_url` more as the default URL to your frontend apps. It could be used to add generic links like:

```
See our privacy policy: ${client_url}/privacy_policy
```

elitan (approved):
LGTM.

I like how this makes it quicker to get started without running into access control issues.

For a future major version (breaking change):

Would it make sense to allow all redirectUrls if only `AUTH_CLIENT_URL` is set?

And we only check URLs if `AUTH_ACCESS_CONTROL_ALLOWED_REDIRECT_URLS` is set?

src/routes/signin/passwordless/email.ts
plmercereau: then `AUTH_CLIENT_URL` could be renamed to `AUTH_DEFAULT_CLIENT_URL` at a later stage

PR #177, opened 2022-06-01
SMS improvements

elitan:
- Able to use both phone number and messaging service id as `from`

This way users can use both a simple phone number without setting up a Twilio messaging service or use a messaging service from Twilio ([doc](https://support.twilio.com/hc/en-us/articles/223181308-Getting-started-with-Messaging-Services)).

- Internal error logging if SMS is unable to be sent


plmercereau (approved):
versioning and changelog are calculated using conventional commits. If we want to keep it that way then commit messages should be structured accordingly.
Or we implement changesets so we manually write changelogs and determine version bumps

Issue #157, opened 2022-05-09
env var to disable sso signup


mmmmillar:
Hi, would you consider an env-var-enabled enhancement that allows SSO login for existing accounts, but disables SSO signup? Atm if you login with Google and a user does not exist it creates a user and logs them in.

eg AUTH_PROVIDER_SIGNUP_ENABLED (true by default)

elitan:
I think this is a good idea.

In the meantime, you can use the `AUTH_ACCESS_CONTROL_*` env vars to block/allow emails to sign up.

PR #184, opened 2022-06-03
feat: disable signup with sso provider

mmmmillar:
This PR allows you to disable SSO signup, while still allowing SSO login for existing accounts (see https://github.com/nhost/hasura-auth/issues/157).

To do this it adds 2 environment variables:
- `AUTH_PROVIDER_SIGNUP_ENABLED` - boolean that defaults to `true` if not set
- `AUTH_PROVIDER_FAILURE_REDIRECT_URL` - string that defaults to `''` if not set

If `AUTH_PROVIDER_SIGNUP_ENABLED` is true, a new user is created and then logged in (as per current functionality)

If `AUTH_PROVIDER_SIGNUP_ENABLED` is false, a new user is not created and the operator is redirected to:
- `AUTH_PROVIDER_FAILURE_REDIRECT_URL?provider={providerName}&failed=true`
- `AUTH_CLIENT_URL?provider={providerName}&failed=true` if `AUTH_PROVIDER_FAILURE_REDIRECT_URL` is not set


elitan:
Thanks for this PR @mmmmillar.

Regarding `AUTH_PROVIDER_SIGNUP_ENABLED`. On a general note, how would this work in practice? If `AUTH_PROVIDER_SIGNUP_ENABLED` is `false` it would never be possible for anyone to sign in with any provider either. 

So would you have `AUTH_PROVIDER_SIGNUP_ENABLED` set to `true` for some time and let users sign in (and sign up), and then set it to `false` so only existing users can sign in? To me, it does not seem very practical. But please let me know your thoughts.

A better solution might be to use ` AUTH_ACCESS_CONTROL_ALLOWED_EMAILS` and ` AUTH_ACCESS_CONTROL_ALLOWED_EMAIL_DOMAINS` to allow only certain email addresses to sign up (and sign in).

mmmmillar:
Hey @elitan, thanks for responding.

### **Regarding: `AUTH_PROVIDER_SIGNUP_ENABLED`**

To answer your question, no I wouldn't set `AUTH_PROVIDER_SIGNUP_ENABLED` to true at all. I would allow sign-up by email/password only, and then sign-in by email/password and SSO.

I guess the best way to describe what I was thinking with this is with a use-case...

I'm working on something at the moment that requires being invited to sign-up. This uses the `sign-in-passwordless-email` functionality to create the user account (and do things like link the user to an org). When the user clicks the link in the email they are redirected to set their password, and then from there they can sign-in with their email and password.

Currently this flow can be broken if a user without an account goes to the sign-in page and clicks on the google (for example) button. This will sign them up right away and sign them in. 

If `AUTH_PROVIDER_SIGNUP_ENABLED` was set to false, the email/password sign-up would stay the same. The difference would be that clicking that google button without having an account would return an error. If a user signed up with email/password first, and then clicked that google button, google would be linked to the existing account and then the user would be signed in.

I can't really achieve something like this atm with `AUTH_ACCESS_CONTROL_ALLOWED_EMAILS` as I don't want to restrict any email addresses if you know what I mean.

### **Regarding: `AUTH_PROVIDER_FAILURE_REDIRECT_URL`**

I was following the guide from http://www.passportjs.org/concepts/authentication/ on verify functions and the different callback scenarios, but can use `sendError` if more appropriate.

elitan:
## Regarding: AUTH_PROVIDER_SIGNUP_ENABLED

Ah, ok. That makes sense. 👍 

I guess you'll still have the same issue with magic link. Anyone can sign up using a magic link. This can be solved, in the future, by disabling the endpoint but overriding it using the admin-secret.

## Regarding: AUTH_PROVIDER_FAILURE_REDIRECT_URL

I think we should use the `sendError` which would make the error flow consistent and we wouldn't have to introduce a separate environment variable.

mmmmillar:
Hey,

Is there anything more I can do with this one? I made the requested changes, probably should have made this clear!

src/utils/env.ts
```diff
@@ -73,6 +73,9 @@ export const ENV = {
   get AUTH_SMS_TWILIO_MESSAGING_SERVICE_ID() {
     return castStringEnv('AUTH_SMS_TWILIO_MESSAGING_SERVICE_ID', '');
   },
+  get AUTH_SMS_TWILIO_FROM() {
```
plmercereau, line 76: could you remove this please? I don't thinks this should be in this PR

src/utils/env.ts
```diff
@@ -133,6 +136,9 @@ export const ENV = {
       locales.push(ENV.AUTH_LOCALE_DEFAULT);
     return locales;
   },
+  get AUTH_PROVIDER_SIGNUP_ENABLED() {
```
plmercereau, line 139: Would you mind adding a short explanation of the environment variable to the [list of env vars](https://github.com/nhost/hasura-auth/blob/main/docs/environment-variables.md)?

plmercereau (changes requested):

src/routes/signin/providers/utils.ts
```diff
@@ -184,6 +190,14 @@ const providerCallback = asyncWrapper(
 
     const user = req.user as UserFieldsFragment;
 
+    if (!user.id) {
+      const failureRedirect = new URL(requestOptions.redirectTo);
+      failureRedirect.searchParams.set('loginFailed', 'true');
+      return sendError(res, 'provider-login-failed', {
+        redirectTo: failureRedirect.toString(),
```
plmercereau, line 197: we are using `signin` instead of `login`. Could you rename to `provider-signin-failed`?

plmercereau: If there is a means to know sign in failed because Oauth sign-up is disabled, that would be great to adapt the error to provide more information to the frontend about why it failed

src/utils/env.ts
mmmmillar: done

mmmmillar:
@plmercereau I think it's ready for review

PR #186, opened 2022-06-06
enabled change password to work with ticket

Svarto:
This replaces my previous PR #183 as this will replicate nhost v1 functionality but also be a lot simpler.

I modified the `/auth/user/password` to also work with a ticket, not only when logged in (but can be logged in AND have a ticket, this is fine too). This allows for a reset email to be sent with a ticket, customize the way it is handled (e.g. deep linking into a mobile app), then use that ticket to change a users password and then log in the user.

There are minimal edits, just adapting the already existing `/auth/user/password` route. Having looked at the `nhost-js` SDK, I also believe it will need only minor edits to the existing `nhost.auth.changePassword` function to allow for another optional variable `ticket`, and this makes that SDK function work both logged in and logged out (assuming you have a valid ticket).

Please let me know your thoughts nhost team (@elitan), if you agree with this approach I can raise another PR to modify the `nhost-js` SDK.


elitan (changes requested):

elitan:
Maybe we should also place this endpoint under `/alpha/users/password` for a while to make sure it works as intended in production and people can incrementally adopt it.

Svarto:
@elitan thanks for the review! I just pushed another commit addressing your feedback.

Please note, I adapted the getter function `getUserByTicket` to return null if the ticket string is empty. This was to handle cases when ticket was not supplied to the request, so it is undefined, but still wanting to handle the getUserByTicket through the function (Typescript).

Regarding your point that it doesn't check for expired tickets. The function actually does here https://github.com/nhost/hasura-auth/blob/856640128e06731b51b23875221e3d77df11b446/src/utils/user/getters.ts#L90 which is how it is used, but the graphql statement you linked to does not. I can link the two, but as I am external contributor I didn't want to mess anything up so left it for now :)

src/routes/user/password.ts
```diff
@@ -7,23 +7,47 @@ import { Joi, password } from '@/validation';
...
+            },
+            {
+              ticketExpiresAt: {
+                _gt: new Date(),
+              },
+            },
+          ],
+        },
+      })
+      .then((gqlRes) => gqlRes.users[0]);
```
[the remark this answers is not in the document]
Svarto: This has been addressed in latest commit.

elitan:
The `ticket` is not sent to the client after the user clicks the link in the email.

https://github.com/nhost/hasura-auth/blob/856640128e06731b51b23875221e3d77df11b446/src/routes/verify/verify.ts#L103-L107

I assume that you would not use this `/verify` endpoint? Instead, you'll use the email variable `ticket` (and either the `redirectTo` or `clientUrl` email variable) directly in your password reset email to construct the link?

src/routes/user/password.ts
```diff
@@ -1,29 +1,36 @@
...
-  { newPassword: string }
+  { newPassword: string, ticket?: string }
 > = async (req, res) => {
-  // check if user is logged in
-  if (!req.auth?.userId) {
+
+  const { ticket } = req.body;
+
+  // get the user from the ticket, but if no ticket then return null
+  const userByTicket = await getUserByTicket(ticket || "")
```
elitan, line 22: I think you can simply do the following here:

```js
const userByTicket = await getUserByTicket(ticket)
```

src/utils/user/getters.ts
```diff
@@ -76,6 +76,12 @@ export const getUserByEmail = async (email: string) => {
 };
 
 export const getUserByTicket = async (ticket: string) => {
+
+  // return null if ticket is empty
+  if (ticket === "") {
+    return null;
+  }
+
```
elitan, line 84: and add `undefined` to the parameter type here and check if `ticket` is set:

```js
export const getUserByTicket = async (ticket: string | undefined) => {

if (!ticket) {
  return null;
}
```

elitan (changes requested):

elitan:
Could you also add tests? 🙏

Svarto:
> The `ticket` is not sent to the client after the user clicks the link in the email.
> 
> https://github.com/nhost/hasura-auth/blob/856640128e06731b51b23875221e3d77df11b446/src/routes/verify/verify.ts#L103-L107
> 
> I assume that you would **not** use this `/verify` endpoint? Instead, you'll use the email variable `ticket` (and either the `redirectTo` or `clientUrl` email variable) directly in your password reset email to construct the link?

Correct, the verify route is not being used at all. This only adds the option to use a ticket with the `nhost.auth.changePassword` function to change password, instead of having to be logged in. The email reset flow would then be:

1. The user enters their email and request their password to be reset on the mobile client through `nhost.auth.resetPassword({ email })`
2. The reset password email template returns a deep link that directs the user back into the mobile app including the ticket, e.g. `yumee://?ticket={{ticket}}&email={{email}}`. You expose "ticket" as a dynamic variable in email templates engine, so this will work.
3. The mobile app takes that deeplink and the ticket, and then through their own custom UX fires a `nhost.auth.changePassword({newPassword: "helloworld", ticket: "email-template-ticket"})`
4. The mobile app can then also automatically log in the user, as we also have the email from the deeplink and the new password.

src/utils/user/getters.ts
Svarto: done!

elitan (approved):

Svarto:
@plmercereau anything that I can do to help you with your review?

plmercereau:
Hello @Svarto, I'll have a look at it today.
In the meantime, would you mind:
- [adjusting the reset password diagram](https://github.com/nhost/hasura-auth/blob/main/docs/workflows/reset-password.md)
- adding a commit starting with `feat: ` with the description of the new feature, so the version bump and the changelog will be automatically generated when we'll publish the next release? See https://github.com/nhost/hasura-auth/blob/main/DEVELOPERS.md#commit-changes

Thanks!

src/routes/user/password.ts
```diff
@@ -1,29 +1,45 @@
...
   {},
-  { newPassword: string }
+  { newPassword: string; ticket?: string }
 > = async (req, res) => {
-  // check if user is logged in
-  if (!req.auth?.userId) {
-    return sendError(res, 'unauthenticated-user');
+  const { ticket } = req.body;
+
+  // get the user from the ticket, but if no ticket then return null
```
plmercereau, line 20: Performance: when using a ticket, the user is fetched twice, and updated twice. That's 4 operations. We could improve this in reviewing how conditions are nested. Something like this draft
```js
let user
if (ticket) {
  user = await getByTicket(ticket)
} else {
  if (!req.auth?.userId) { return sendError('...') }
  user = (await gqlSdk.user({ id: userId })).user
}

if (!user) { return sendError('...') }
const { newPassword } = req.body;
const passwordHash = await hashPassword(newPassword);

await gqlSdk.updateUser({
    id: user.id,
    user: {
      passwordHash,
      ticket: ticket ? null : undefined // not sure on this one, but I think Hasura undestands `undefined` as "do not change the value"
    },
  });
  return res.send(ReasonPhrases.OK);
};
```

src/utils/user/getters.ts
```diff
@@ -75,7 +75,13 @@ export const getUserByEmail = async (email: string) => {
   return users[0];
 };
 
-export const getUserByTicket = async (ticket: string) => {
+export const getUserByTicket = async (ticket: string | undefined) => {
```
plmercereau, line 78: this change wouldn't be necessary anymore with the previous comment

plmercereau (changes requested):

src/routes/user/password.ts
plmercereau: we could even use one single mutation like "update user.passwordHash, user.ticket where valid ticket OR id" and send an error if no user has been found:

```ts
const { ticket } = req.body;
  const userId = req.auth?.userId;
  if (!userId && !ticket) {
    return sendError(res, 'unauthenticated-user');
  }

  const { updateUsers } = await gqlSdk.updateUserWhere({
    where: {
      _or: [
        {
          _and: [
            {
              ticket: {
                _eq: ticket,
              },
            },
            {
              ticketExpiresAt: {
                _gt: new Date(),
              },
            },
          ],
        },
        { id: { _eq: userId } },
      ],
    },
    user: {
      passwordHash: await hashPassword(req.body.newPassword),
      ticket: ticket ? null : undefined,
    },
  });

  if (!updateUsers?.affected_rows) {
    return sendError(res, 'invalid-user-ticket'); // Invalid user or verification ticket
  }

  return res.send(ReasonPhrases.OK);
```

src/utils/user/getters.ts
Svarto: Done!

src/routes/user/password.ts
Svarto: thanks a lot for the review @plmercereau !

I went with your first suggestion, as I believe the added error granularity will be welcomed. I also confirmed that Hasura does not update a field if "undefined" is passed in as its value.

Have a look at the latest commit status and let me know what you think. Apologies for the duplicative commit, I got out of sync with the remote repository and had to merge the remote changes after I had already locally committed my own edits.

Svarto:
@plmercereau Should all be done now. Also updated the change-password diagram as the PR impacts that workflow as well.

plmercereau:
> @plmercereau Should all be done now. Also updated the change-password diagram as the PR impacts that workflow as well.

Awesome! I will have a look tomorrow. Many thanks!

plmercereau (approved):

PR #188, opened 2022-06-07
fix: add `email` and `newEmail`  variables to all email templates

plmercereau:
Email template "locals" are now strongly typed to avoid such a problem in the future.

I'm not convinced we should add `newEmail` everywhere as it only has a meaning when the user tries to change their email.

### DIFF
--- a/src/email.ts
+++ b/src/email.ts
@@ -113,10 +113,22 @@ const readRemoteTemplate = async (
   }
 };
 
+type EmailOptions = {
+  link: string;
+  displayName: string;
+  email: string;
+  newEmail: string;
+  ticket: string;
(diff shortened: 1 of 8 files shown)


elitan:
> I'm not convinced we should add `newEmail` everywhere as it only has a meaning when the user tries to change their email.

Kind of agree, but instead of us being selective of the variables, we are consistent and always make them available. Then it's up to the developer how they want to use the variables in the email templates.

From my perspective, it does not hurt to add extra variables, and also it becomes easier to document because we can just say: "these variables are always available on email templates" instead of listing individual email templates and what variables are available.

elitan (approved):

PR #191, opened 2022-06-09
fix: workaround for outlook safe links

mmmmillar:
This PR adds a workaround for Outlook safe links (see https://github.com/nhost/hasura-auth/issues/189).

Tested working.

PR #192, opened 2022-06-10
fix: do not delete ticket, but wait for it to expire

quentin-decre:
I have used the password less authentication quite a lot lately.
And I have came to have users with (I guess) anti-spam software, which opens the link automatically.
As Hasura Auth deletes the ticket on the first opening, the user can never authenticate with this method.

So I removed the ticket deletion part. 
I have checked that Hasura Auth checks that the token is not expired, so it doesn't seem to be a security issue.

Maybe we could reduce the expireAt to now() + 5 / 10 minutes instead ?

### DIFF
--- a/src/routes/verify/verify.ts
+++ b/src/routes/verify/verify.ts
@@ -53,13 +53,8 @@ export const verifyHandler: RequestHandler<
     return sendError(res, 'invalid-ticket', { redirectTo });
   }
 
-  // user found, delete current ticket
-  await gqlSdk.updateUser({
-    id: user.id,
-    user: {
-      ticket: null,
-    },
-  });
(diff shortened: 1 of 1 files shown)


mmmmillar:
This may be related to https://github.com/nhost/hasura-auth/pull/191

plmercereau:
> This may be related to #191

Yes it is most likely related to the above PR. @quentin-decre could you check if solves your issue? I hope it does as I feel we should find ways to avoid removing security measures when possible

plmercereau:
Hi @quentin-decre,
Could you confirm #191 solved your problem? In the meantime I'm closing this PR as #191 has been released in [0.9.1](https://github.com/nhost/hasura-auth/releases/tag/v0.9.1)

PR #195, opened 2022-06-11
fix: do not actually follow redirection when redirectTo is invalid

plmercereau:
(except for GET requests: /verify and social providers)

let's take this react-like code:
```jsx
const { signUpEmailPassword, needsEmailVerification } = useSignUpEmailPassword()
const handle = () => {
  signUpEmailPassword({
    email: 'bob@thebuilder.com',
    password: 'strong-password',
    options: {
      redirectTo: 'http://unauthorised-url.com'
    } 
  })
}
```
Instead of returning an http `400`, the server would actually forward the error to `AUTH_CLIENT_URL`.

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -149,13 +149,13 @@ export const sendError = (
     customMessage,
     redirectTo,
   }: { customMessage?: string; redirectTo?: string } = {},
-  redirect?: boolean
+  forwardRedirection?: boolean
 ) => {
   const error = ERRORS[code];
   const message = customMessage ?? error.message;
   const status = error.status;
 
-  if (redirect && redirectTo) {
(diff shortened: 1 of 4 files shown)


elitan:
I'm not following exactly what the original behavior was and what the new behavior is. Is this a bug fix or something else?

plmercereau:
yes
when the redirectTo fo the above code is incorrect, it actually follows the `AUTH_CLIENT_URL` redirection and adds an `error` and `errorDescription` to the url. The expected behaviour is not to follow anything and to return a `400` code.
Errors should only be redirected when using the `/verify` endpoint, or after the oauth callback

elitan (approved):

PR #198, opened 2022-06-17
feat(auth): FIDO2 Webauthn authentication

### LINKED ISSUE #153: Webauthn Yubikey
Suggested by @masterbater

L3K0V:
close #153


L3K0V:
@plmercereau @elitan I have added webauthn FIDO2 support for signing up and in. I wonder if `webauthn` is a good name for the endpoints. Do you think `passkey` is a better one choice?

plmercereau:
Nice!
`webauthn` seems a good name as it is the name of the standard

I haven't looked at your PR closely yet, but that's great to see you already used our internals already 👌 (codegen, openapi, conventional commits). A couple of comments though:
- I seems you're adding a `user_authenticators` table, but I don't see any SQL SDL in the `migrations` folder, would you mind adding it?
- Could you consider adding tests as well?

👍

L3K0V:
> I seems you're adding a user_authenticators table, but I don't see any SQL SDL in the migrations folder, would you mind adding it?

Yeah, sure it's here but git does not track it for some reason, maybe global ignore configuration...

> Could you consider adding tests as well?

Yes, I will do my best, still there's also some configurations to be added and ENV variables

src/routes/signup/webauthn.ts
```diff
@@ -0,0 +1,163 @@
...
+import {
+  generateRegistrationOptions,
+  verifyRegistrationResponse,
+} from '@simplewebauthn/server';
+import { email, Joi, passwordInsert, registrationOptions } from '@/validation';
+import { createUserAndSendVerificationEmail } from '@/utils/user/email-verification';
+
+export const signUpWebauthnSchema = Joi.object({
+  email: email.required(),
+  password: passwordInsert.optional(),
```
plmercereau, line 15: why adding password as an option? It is not required by webauthn (and we can also assume users opting for wbeauthn don't want any password).

They can eventually add a password after sign-up

src/routes/signin/webauthn.ts
```diff
@@ -0,0 +1,158 @@
...
+    email: string;
+  }
+> = async (req, res) => {
+  const { body } = req;
+  const { email } = body;
+
+  const user = await getUserByEmail(email);
+
+  if (!user) {
+    return sendError(res, 'invalid-email-password');
```
plmercereau, line 38: probably better to use `user-not-found` instead

src/routes/signup/webauthn.ts
L3K0V: I was wondering if this is a nice optional feature or should be removed, so you propose not to have it?

plmercereau: I think that is better not to have it. Users can set their password at a later stage in using the `/user/password` endpoint.
We are doing that way for Oauth or passwordless email, I would keep the same logic

L3K0V:
@plmercereau one think to discuss and maybe help me with. I know Webauthn can be also used as a TFA, right?
Can you give me some hints around the TFA flow and maybe some guidelines how we can add webauthn there.

plmercereau:
> 



> @plmercereau one think to discuss and maybe help me with. I know Webauthn can be also used as a TFA, right? Can you give me some hints around the TFA flow and maybe some guidelines how we can add webauthn there.

For the moment the only 2FA we implemented is email/password + TOTP.
I feel webauthn as a second FA would be a really awesome addition, but I suggest we keep this feature out of the scope of this PR so we don't jump into too much complexity, as we also need to have a deeper understanding on which first and second factors we want to support - and that is a discussion that goes beyond webauthn

src/routes/signin/webauthn.ts
L3K0V: Done ✅

src/routes/signup/webauthn.ts
L3K0V: Changed.

L3K0V:
@plmercereau we can move this for review. I have added tests and env variables.

About the tests, I'm having troubles with jest mocking the `verifyAuthenticationResponse` and `verifyRegistrationResponse` from the `@simplewebauthn` library. Can you help me with that, so I can cover the rest of the flows?

L3K0V:
@plmercereau is there anything I can help about the PR?

plmercereau:
Hello @L3K0V, I'm currently having a look at it, will keep you posted very soon :)
Thanks for your patience, this PR is really nice

src/routes/signup/webauthn.ts
```diff
@@ -0,0 +1,170 @@
...
+  });
+
+  return res.send(registrationOptions);
+};
+
+export const signUpVerifyWebauthnHandler: RequestHandler<
+  {},
+  {},
+  {
+    credential: any;
```
plmercereau, line 83:
```suggestion
    credential: RegistrationCredentialJSON;
```

src/routes/signin/webauthn.ts
```diff
@@ -0,0 +1,164 @@
...
+  });
+
+  return res.send(options);
+};
+
+export const signInVerifyWebauthnHandler: RequestHandler<
+  {},
+  {},
+  {
+    credential: any;
```
plmercereau, line 76:
```suggestion
    credential: RegistrationCredentialJSON;
```

src/routes/signin/webauthn.ts
```diff
@@ -0,0 +1,164 @@
+import { sendError } from '@/errors';
+import { ENV, getSignInResponse, getUserByEmail, gqlSdk } from '@/utils';
+import { RequestHandler } from 'express';
+
+import {
+  generateAuthenticationOptions,
+  verifyAuthenticationResponse,
+} from '@simplewebauthn/server';
+import { email, Joi } from '@/validation';
```
plmercereau, line 9:
```suggestion
import { RegistrationCredentialJSON} from '@simplewebauthn/typescript-types';
import { email, Joi } from '@/validation';
```

src/routes/signup/webauthn.ts
```diff
@@ -0,0 +1,170 @@
+import { sendError } from '@/errors';
+import { UserRegistrationOptions } from '@/types';
+import { ENV, getSignInResponse, getUserByEmail, gqlSdk } from '@/utils';
+import { RequestHandler } from 'express';
+
+import {
+  generateRegistrationOptions,
+  verifyRegistrationResponse,
+} from '@simplewebauthn/server';
+import { email, Joi, registrationOptions } from '@/validation';
```
plmercereau, line 10:
```suggestion
import { RegistrationCredentialJSON} from '@simplewebauthn/typescript-types';
import { email, Joi, registrationOptions } from '@/validation';
```

src/utils/user/email-verification.ts
```diff
@@ -0,0 +1,138 @@
...
+import { getUserByEmail } from './getters';
+
+const sendEmailIfNotVerified = async (
+  email: string,
+  newEmail: string,
+  user: any,
+  displayName: string,
+  ticket: string,
+  redirectTo: string
+) => {
```
plmercereau, line 19: We can improve the readability and maintenance in wrapping params in an object:

```suggestion
import { UserQuery } from '../__generated__/graphql-request';

const sendEmailIfNotVerified = async ({
  email,
  newEmail,
  user,
  displayName,
  ticket,
  redirectTo,
}: {
  email: string;
  newEmail: string;
  user: NonNullable<UserQuery['user']>;
  displayName: string;
  ticket: string;
  redirectTo: string;
}) => {```

File: src/utils/user/email-verification.ts [hunk 8]
```diff
@@ -0,0 +1,138 @@
...
+      email,
+      email,
+      existingUser,
+      displayName,
+      existingUser.ticket!,
+      redirectTo
+    );
+
+    return existingUser;
+  }
```
plmercereau, line 100:
```suggestion
  if (!existingUser.ticket) {
    throw Error(`No ticket found for the user ${existingUser.id}`);
  }

  if (existingUser) {
    await sendEmailIfNotVerified({
      email,
      newEmail: email,
      user: existingUser,
      displayName,
      ticket: existingUser.ticket,
      redirectTo,
    });
```

src/utils/user/email-verification.ts
```diff
@@ -0,0 +1,138 @@
...
+  });
+
+  await sendEmailIfNotVerified(
+    email,
+    user.newEmail,
+    user,
+    displayName,
+    ticket,
+    redirectTo
+  );
```
plmercereau, line 135:
```suggestion
  await sendEmailIfNotVerified({
    email,
    newEmail: user.newEmail,
    user,
    displayName,
    ticket,
    redirectTo,
  });
```

L3K0V:
@plmercereau this is adjusted sample from `@simplewebauthn` to test the implementation. Notice in the HTML once you are registered, you should change the hardcoded user id with the newly created one.

[webauthn.zip](https://github.com/nhost/hasura-auth/files/9019463/webauthn.zip)

Let me know if you have idea how to mock the `@simplewebauthn` internals in order to test the rest of the functionality.

plmercereau (changes requested):

src/routes/signin/index.ts
```diff
@@ -83,6 +89,36 @@ router.post(
...
 
+/**
+ * POST /signin/webauthn
+ * @summary Sign in using email via FIDO2 Webauthn authentication
+ * @param {SignInWebauthnSchema} request.body.required
+ * @return {InvalidRequestError} 400 - The payload is invalid - application/json
+ * @return {DisabledEndpointError} 404 - The feature is not activated - application/json
+ * @tags Authentication
+ */
+router.post(
```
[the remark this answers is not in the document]
L3K0V: Done, please check  if they are sufficient

src/env-vars-check.ts
```diff
@@ -191,6 +191,20 @@ if (ENV.AUTH_SMS_PROVIDER) {
   }
 }
 
+if (ENV.AUTH_WEBAUTHN_ENABLED) {
```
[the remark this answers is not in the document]
L3K0V: Done

migrations/00005_setup-webauthn.sql
```diff
@@ -0,0 +1,20 @@
+-- start a transaction
```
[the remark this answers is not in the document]
L3K0V: Done

L3K0V:
@plmercereau something strange happens on the tests now... 🤔 

Before each test I'm deleting all users and there is just one single test failing.
When I call `webauth/signup` I check if there existing user with the email. Should not have as I delete before the tests the users, but all the time it gives me that in the DB there is a user with this email.

image

Could you be so kind to check and help me?

L3K0V:
@plmercereau done, let me know if there's anything

L3K0V:
@plmercereau moving this to draft as I found a flaw. 

I allow user to sign up multiple times with same email but with different devices or browsers, making possible to hijack account by knowing the email. 

Adding additional authenticators aka. (device, browser) should be done differently. My idea is to use ticket, clicking the email to be redirected to the app giving a new challenge to be signed and passed back to the backend. Need to check and think about a bit.

If you have some ideas, they will be more than welcome. 🙏

plmercereau:
> @plmercereau moving this to draft as I found a flaw.
> 
> I allow user to sign up multiple times with same email but with different devices or browsers, making possible to hijack account by knowing their email.
> 
> Adding additional authenticators aka. (device, browser) should be done differently. My idea is to use ticket, clicking the email to be redirected to the app giving a new challenge to be signed and passed back to the backend. Need to check and think about a bit.
> 
> If you have some ideas, they will be more than welcome. 🙏

Not behind my laptop to think this further, but what about signing up with one authenticator, then requiring the JWT to be send in the headers to enable another one?

L3K0V:
> Not behind my laptop to think this further, but what about signing up with one authenticator, then requiring the JWT to be sent in the headers to enable another one?

This will help, but will not cover one very important and common case - User want to sign in from new device, where there's no active session. Imagine I sign up and login on the browser and later on I want to sign in from my phone.

L3K0V:
@plmercereau sorry again for bothering...

Good news! I was able to configure the Webauthn implementation to support Passkeys. Passkeys are synced between devices. Safari support passkeys, later this year Android devices as well. https://www.passkeys.io

Now on sign-up, if user is verified and already have authenticator setup, will require Bearer token as you proposed. 

I suggest to proceed here with this PR, and maybe open new one soon about signing in from new device without a session and without passkey support.

plmercereau:
Hello @L3K0V,
I still see a major issue for now: anyone can use the email of an existing user to authenticate, when this use has not yet set a webauthn authenticator.

- I think it is important we don't mix terms: the meaning of "sign-up" should remain "a user gets registered to the application". In the current `/signup/webauthn` endpoint, it is used for both "a user is getting registered" and "a registered user is activating webauthn".
- We are somehow not starting with the easiest use case due to hasura-auth design: there is no concept of "username" (yet), and an email is both considered both a user identifier and a means to authenticate (hence the need to verify emails). With the following signup endpoint, it will be easier to let users register with webauthn with a username (the payload would then be either an email or a username)
- Still because we're mixing usernames and emails, we are missing one important point: once authenticated by any means (webauthn, magic link, email + password, oauth...), a user should be able to add a new device. Hence the following `activate` endpoint.

I will gather my thoughts and come with some refactoring on your PR, but here's the idea:

1. `/signup/webauthn`. Payload: email + credential (and at a later stage, also possible: username + credential)
   - if email verification is enabled, return an error
   - if the email already exists, return an error
   - else add authenticator and return refresh+access tokens

2. `/user/webauthn/activate`. Payload: credential + bearer token
   - if not authenticated, return an error
   - else add authenticator

3. `/signin/webauthn`. Payload: email + credential
   - no change in the logic

I understand it makes things quite restrictive (until we implement user names), but the approach is safer.
Let's take some examples:
- when email verification is enabled:
   1. the user signs up with magic link. Once they clicked on the email link, they are authenticated
   2. they activate webauthn for their current device
   3. sign out, sign in with webauthn
- when email verification is enabled:
   1. the user signs up using `/signup/webauthn`. Webauthn is activated for their current device
   2. sign out, sign in with webauthn
- a user signed up with their email and activated webauthn on device A
   1. they sign in using a magic link / email + password on device B
   2. they activate webauthn on device B
   - note 1:. when email links are not involved, it is possible to chain 1 and 2 on the frontend so the user will do both steps all together
   - note 2: when using email links, it is possible to redirect the user to a page that will automatically open the webauthn dialog 

I'll be quite on/off until Monday, but I'll dig this further at the beginning of the week - what you've done so far is really good and I am convinced this will become a key hasura-auth feature.

In the meantime, I'm curious to get your thoughts about this message.

L3K0V:
Hey @plmercereau,

Your proposal sounds like a plan and I will do the necessary adjustments.

Can you just be more specific on these, I did not get them:

> - when email verification is enabled:
the user signs up with magic link. Once they clicked on the email link, they are authenticated
they activate webauthn for their current device
sign out
sign in with webauthn
> - when email verification is enabled:
the user signs up using /signup/webauthn. Webauthn is activated for their current device
sign out
sign in with webauthn

is the second point really again email verification enabled, or should be disabled?

> note 1:. when email links are not involved, it is possible to chain 1 and 2 on the frontend so the user will do both steps all together

by 1 and 2 you meant?

## Overall

Implementing these as you mentioned is more restrictive, which is also some kind a good thing, but adding a new device, or device loss, requires additional way of login before adding new authenticator (which btw is the main downside of webauthn by definition). If user uses passkeys the sign in can survive between different devices or a device loss. Unfortunately not using passkeys, webauthn sounds to me more like TFA **in some cases**, but I think we can have a great initial implementation, covering most of the cases and not compromising the security.

Any thoughts?

plmercereau:
> is the second point really again email verification enabled, or should be disabled?

Sorry, I meant "When email verification is **disabled**". I corrected

> by 1 and 2 you meant?

Gave an example, but the frontend code could look like this:
```js
const { session } = await nhost.auth.signIn({email, password})
if (session) {
  await nhost.auth.activateWebauthn()
}
```

> Implementing these as you mentioned is more restrictive, which is also some kind a good thing, but adding a new device, or device loss, requires additional way of login before adding new authenticator (which btw is the main downside of webauthn by definition). If user uses passkeys the sign in can survive between different devices or a device loss. Unfortunately not using passkeys, webauthn sounds to me more like TFA **in some cases**, but I think we can have a great initial implementation, covering most of the cases and not compromising the security.

I don't see any other option: before adding a device to the user, we need to know the user is the one he's pretending to be i.e. he needs to be authenticated on this device. The above design will work with any authentication method, but yes, at least  one other authentication method should be available: email+password, magic link, oauth...

L3K0V:
@plmercereau when you have time can you check? I think I'm ready.

.env.example
```diff
@@ -71,6 +71,12 @@ AUTH_PROVIDER_STRAVA_ENABLED=
 AUTH_PROVIDER_STRAVA_CLIENT_ID=
 AUTH_PROVIDER_STRAVA_CLIENT_SECRET=
 
+# WEBAUTHN
+AUTH_WEBAUTHN_ENABLED=
+AUTH_WEBAUTHN_RP_ID=localhost
+AUTH_WEBAUTHN_RP_NAME=Nhost App
```
plmercereau, line 77:
```suggestion
AUTH_WEBAUTHN_RP_NAME='Nhost App'
```

docs/configuration.md
```diff
@@ -93,14 +93,33 @@ It is possible to add a step to authentication with email and password authentic
...
+Hasura Auth supports [Webauthn authentication](https://en.wikipedia.org/wiki/WebAuthn). Users can sign up and sign in using different strong authenticators like Face ID, Touch ID, Fingerprint, Hello Windows etc. using supported devices. **Passkeys are supported for cross-device sign in.**
+
+**Each user can sign up only once using webauthn. Existing users can add subsequent webauthn authenticators (new device or browser) via `/user/webauthn/add`, which requires Bearer authentication token.**
+
+Enabling and configuring of the Webauthn can be done by setting these env variables:
+
+```bash
+AUTH_WEBAUTHN_ENABLED=true
+AUTH_WEBAUTHN_RP_ID=https://my-app.vercel.com
+AUTH_WEBAUTHN_RP_NAME=My App
```
plmercereau, line 117:
```suggestion
AUTH_WEBAUTHN_RP_NAME='My App'
```

plmercereau (approved):

plmercereau:
Bring it on!

PR #210, opened 2022-07-05
feat(passwordless-sms): support for templates for the sms message

L3K0V:


L3K0V:
@plmercereau can you check this one as well?

src/env-vars-check.ts
```diff
@@ -174,6 +174,7 @@ if (PROVIDERS.github) {
 if (ENV.AUTH_SMS_PROVIDER) {
   if (ENV.AUTH_SMS_PROVIDER === 'twilio') {
     [
+      'APP_NAME',
```
plmercereau, line 177: this introduces a breaking change: SMS authentication won't work in upgrading to the new version without setting a value to this new `APP_NAME` environment variable.
I would make sure what's implemented does not introduce any regression

plmercereau (changes requested):

src/env-vars-check.ts
L3K0V: `APP_NAME` env variable is needed in my opinion for serving many purposes like emails, sms, webauthn relying party name and maybe many more. I wonder why there wasn't  one till one 🤔 I assumed that such variable easily can be available in Nhost. 

Rather introducing breaking change we can fallback to some default string, but I'm not sure if it's a good idea, imagine:

`123456 is your verification code for hasura-auth`

Let me know if there's something to adjust here.

L3K0V:
@plmercereau I have merged your updated here, so let me know if there's something else I can help to get this rolling

src/routes/signin/passwordless/sms/sms.ts
```diff
@@ -91,8 +92,16 @@ export const signInPasswordlessSmsHandler: RequestHandler<
           to: phoneNumber,
         });
     } else {
+      const template = 'signin-passwordless-sms';
```
plmercereau, line 95: this is somehow a "breaking" change: when an app upgrades to this version, it will use the default template that will render the following sms:
`123456 is your verification code for undefined.`
I would change the default template back to
`Your code is ${code}` and `Votre code est ${code}` in french.

We could then change the default template to your suggestion when we (nhost) will have implemented `APP_NAME` in our console

L3K0V: Sorry @plmercereau is this for me?

Do you want me to patch this? When we do not have `APP_NAME` a fallback to be introduced?

L3K0V: Or is a general discussion?

email-templates/en/signin-passwordless-sms/body.txt
```diff
@@ -0,0 +1 @@
+${code} is your verification code for ${appName}.
```
plmercereau, line 1:
```suggestion
Your code is ${code}.
```

email-templates/fr/signin-passwordless-sms/body.txt
```diff
@@ -0,0 +1 @@
+${code} est votre code de validation pour ${appName}.
```
plmercereau, line 1:
```suggestion
Votre code est ${code}.
```

plmercereau (changes requested):

src/routes/signin/passwordless/sms/sms.ts
plmercereau: Applications using the latest released version of hasura-auth have codes like this:
```
Your code is 123456
```

When upgrading with the same configuration, codes will look like this:

```
123456 is your verification code for .
```

Please see the following suggestions

L3K0V: Yes I already updated them. Any ideas how we can include APP name to the templates? It will be a breaking change in all cases.

plmercereau (approved):

elitan (changes requested):
This PR tries to do too much IMO.

1. Adding SMS templates 👍 
2. Changing default values for env vars
3. Adding env var(s) with no direct association with SMS templates.

I think this PR should only do 1. SMS templates and revert all other unrelated changes.

L3K0V:
> This PR tries to do too much IMO.
> 
> 1. Adding SMS templates 👍
> 2. Changing default values for env vars
> 3. Adding env var(s) with no direct association with SMS templates.
> 
> I think this PR should only do 1. SMS templates and revert all other unrelated changes.

That's why I reuse the current email templating implementation and introduce SMS templates. The only break change I have introduced was the `APP_NAME` which we found a workaround for and was applied here https://github.com/nhost/hasura-auth/pull/210/commits/346a5106f3fe9f4b7e8e0954c3647897387c0b3b.  @plmercereau introduce a lot new adjustments for `unifying of the templates` in https://github.com/dodi-smart/hasura-auth/pull/2

Shall I open new PR and apply just SMS templates?

PR #212, opened 2022-07-08
fix: remove comment on the auth schema

plmercereau:
The Postgres user that runs the user should be able to run the migrations even if they are not the owner of the auth schema.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ get-version:  ## Return version.
 
 .PHONY: dev
 dev: check-port install dev-env-up  ## Start development environment.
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -24,7 +24,7 @@ get-version:  ## Return version.
 
 .PHONY: dev
 dev: check-port install dev-env-up  ## Start development environment.
-	bash -c "trap 'make dev-env-down' EXIT; pnpm dev"
```
plmercereau, line 27: only start hasura-auth, not graphql-codegen, that can be run manually when required

nunopato (approved):

PR #224, opened 2022-09-01
Improve WebAuthn

plmercereau:
### Broaden WebAuthn authenticators & algorithms
Only some authenticators were allowed in the initial configuration. The service now accepts by default any authenticator that is available in the client. For instance, in Mac + Chrome, it was only possible to use Touch ID, but not an Android device linked to this account.

### Add an optional `nickname` to security keys
It is common for the users to identify a security key by a nickname, so they have a human-readable name to list/remove them. See GitHub implementation of WebAuthn.

### Remove webauthn signup endpoints
The `/signup/webauthn` presented no added value as it can be achieved with running this sequence:
1. sign up (not only with email+password but with any method that gets the user email)
2. add a security key

### OpenApi specs
Generate improvements in the OpenApi specs


src/routes/signin/webauthn.ts
```diff
@@ -119,15 +117,15 @@ export const signInVerifyWebauthnHandler: RequestHandler<
...
-    .then((gqlres) => gqlres.authUserAuthenticators);
-
-  const authenticator = userAuthenticators.find(
-    (auth) => auth.credentialId === credential.id
-  );
+    .then(({ authUserAuthenticators }) =>
+      authUserAuthenticators.find(
+        ({ credentialId }) => credentialId === credential.id
+      )
+    );
```
elitan, line 128: I would try not to do anything in callback functions (`.then()`) and instead always follow the same `async`/`await` pattern.

Something like this:

```js
const { authUserAuthenticators } = await gqlSdk
  .getUserAuthenticators({
    id: user.id,
  })

const authenticator = authUserAuthenticators.find(
  (auth) => auth.credentialId === credential.id
);

File: src/routes/signin/webauthn.ts [hunk 2]
```diff
@@ -47,17 +47,15 @@ export const signInWebauthnHandler: RequestHandler<
...
   }
 
-  const userAuthenticators = await gqlSdk
-    .getUserAuthenticators({
-      id: user.id,
-    })
-    .then((gqlres) => gqlres.authUserAuthenticators);
+  const { authUserAuthenticators } = await gqlSdk.getUserAuthenticators({
+    id: user.id,
+  });
```
elitan, line 52: No callbacks, good! 👍

docs/workflows/webauthn.md
```diff
@@ -1,44 +1,36 @@
-# Sign up and sign in users with Webauthn using strong authenticator
+# Security Keys with WebAuthn
 
-User can sign up with webauthn only if email verification is disabled. When email verification is enabled, user must verify it's email, login via password, magic link or ther credentials and then add webauthn authenticator via `/user/webauthn/add` endpoint instead. After that user can sign in using the webauthn authenticator.
+Hasura-auth implements the WebAuthn protocol to sign in with security keys.
 
-## Sign up
+A user needs first to sign up with another method, for instance email+password, passwordless email or Oauth, then to add their security key to their account.
+
+Once the security key is added, it is then possible to sign in with it, using the email as a user name.
```
elitan, line 7: "user name" => "username"

docs/workflows/webauthn.md
```diff
@@ -1,44 +1,36 @@
...
+
+## Add a security key
+
+Users can add multiple authenticators, for example when they need to login from multiple devices or browsers. To do that, they should have a valid session.
```
elitan, line 11: Is "authenticators" the same thing as "security keys"? 

Asking because we're using the terminology "security keys" in the header for this secion.

elitan: What do you mean by "valid session"? Do you mean users must be signed in when adding a security key?

elitan (changes requested):

docs/workflows/webauthn.md
plmercereau: yes - catching up previous explanations

elitan (approved):

migrations/00007_add_authenticator_nickname.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+ALTER TABLE auth.user_authenticators
+  ADD COLUMN nickname text;
```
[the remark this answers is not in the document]
elitan: I think it's better not to have `IF NOT EXISTS` as it could hide underlying issues with the schema.

The `nickname` column should not exist. If it does, this migration will fail, and the underlaying problem must be fixed.

If a `nickname` column with type `int` existed already, this migration would be successful, but then errors would show elsewhere. 

Better to catch these errors early.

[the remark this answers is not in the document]
nunopato: Because we don't live in a perfect world :) Imagine being able to run all migrations for a broken project and end up at the desired state, versus having to, as an example, get into RDS, and run SQL commands to fix what's broken because users messed with the schema. We do not control what our users do to their databases. But we will have to fix those issues once they ask us to. It is very different having to manage our own database versus having to manage thousands.

PR #225, opened 2022-09-05
fix(webauthn): use the hostname of `AUTH_SERVER_URL` as a default relying party

plmercereau:
Our current WebAuthn implementation was introducing a new required `AUTH_WEBAUTHN_RP_ID` environment variable that contained the relying party. It can however be deduced from `AUTH_SERVER_URL`'s hostname, making this env var optional.

For instance:
```bash
AUTH_SERVER_URL=http://localhost:1337/v1/auth
# deduced:
AUTH_WEBAUTHN_RP_ID=localhost
```

```bash
AUTH_SERVER_URL=https://xyv.nhost.run/v1/auth
# deduced:
AUTH_WEBAUTHN_RP_ID=xyv.nhost.run

### CODE REVIEW

File: src/utils/env.ts [hunk 1]
```diff
@@ -96,7 +96,10 @@ export const ENV = {
     return castBooleanEnv('AUTH_WEBAUTHN_ENABLED', false);
   },
   get AUTH_WEBAUTHN_RP_ID() {
-    return castStringEnv('AUTH_WEBAUTHN_RP_ID', '');
+    return castStringEnv(
+      'AUTH_WEBAUTHN_RP_ID',
+      ENV.AUTH_SERVER_URL && new URL(ENV.AUTH_SERVER_URL).hostname
+    );
```
elitan, line 102: Are there any use cases where uses don't want to deduct the `AUTH_WEBAUTHN_RP_ID` from `AUTH_SERVER_URL`? 

If no, maybe we should remove `AUTH_WEBAUTHN_RP_ID`?

elitan (approved):

PR #227, opened 2022-09-05
fix: adding profile variable for apple sign in

elitan:
Without this variable the `idToken` was empty.


elitan:
Following the example here: https://github.com/ananay/passport-apple-example/blob/master/src/index.js#L31

plmercereau (approved):

plmercereau:
No conventional commit: we mustn't forget to add this one to the next changelog

PR #228, opened 2022-09-06
feat: workos oauth provider

plmercereau:
### Checklist

- [x] No breaking changes
- [x] Tests pass
- [ ] New features have new tests:  no: Oauth


package.json
```diff
@@ -30,6 +30,7 @@
     "@riderize/passport-strava-oauth2": "^1.1.1",
     "@simplewebauthn/server": "^5.3.0",
     "@types/passport-oauth2": "^1.4.11",
+    "@workos-inc/node": "^2.12.0",
```
elitan, line 33: It seems we're not using this package

elitan (changes requested):

package.json
plmercereau: true, it's included as a dependency in `passport-workos`

plmercereau: actually no, we need to install it so `passport-workos` works

elitan (approved):

PR #244, opened 2022-09-20
fix: webauthn signup endpoints

plmercereau:


elitan (approved):
The signup flow is using `users.newEmail` to store the user between step 1 and 2 during signup, which is a bit hacky but ok for now. This is something we should refactor in the future.

Issue #243, opened 2022-09-20
Azure AD integration


yannickglt:
We use Microsoft Azure AD in our company and I would like to use it for Hasura authentication. 

Would you agree if I create a pull request to support Azure AD login using the https://github.com/AzureAD/microsoft-authentication-library-for-js library? Is there a way to chat together if I need help somewhen?

Thanks

elitan:
Thanks for the initiative.

Yes, we would accept a PR like that.

We're on Discord if you need to chat: https://nhost.io/discord

PR #248, opened 2022-09-23
feat(provider): add azure ad provider

yannickglt:
Before submitting this PR:

### Checklist

- [x] No breaking changes
- [x] Tests pass

### Breaking changes

### Tests

### Documentation

### Conventional commits

Versioning and changelog are generated following [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/). Please do your best to follow the convention, but don't worry about it too much. We will be most likely moving away to [changesets](https://github.com/changesets/changesets) soon.

Closes https://github.com/nhost/hasura-auth/issues/243


yannickglt:
**Demo:**

https://user-images.githubusercontent.com/1006426/192337374-d7ff5291-d327-49c4-bf67-158532302bcf.mp4

src/routes/signin/providers/utils.ts
```diff
@@ -255,21 +255,39 @@ export const initProvider = <T extends Strategy>(
           req: RequestWithState<ProviderCallbackQuery, { user: string }>,
           accessToken: string,
           refreshToken: string,
-          idToken: string,
+          params: unknown,
```
elitan, line 258: We must ensure this change does not break the Apple provider.

elitan:
How should we think about Azure AD? Are all Microsoft accounts able to sign in using this, or how does it work? Are there any good resources you can recommend regarding this?

src/routes/signin/providers/utils.ts
yannickglt: [There](https://github.com/nhost/hasura-auth/blob/6b3df9a8870186f6ef270383fa318132e932f4bd/src/routes/signin/providers/utils.ts#L274) the params refers to the token as a string for the Apple integration (as before), while for the new [Azure AD](https://github.com/nhost/hasura-auth/blob/6b3df9a8870186f6ef270383fa318132e932f4bd/src/routes/signin/providers/utils.ts#L283) integration the params is an object containing the token.
IMO it behaves as before but I did not test that part as we don't use Apple SSO internally.

yannickglt:
Azure Active Directory allows connecting through [Outlook, OneDrive, Xbox LIVE, or Microsoft 365](https://learn.microsoft.com/en-us/azure/active-directory/fundamentals/active-directory-whatis). 

Depending on the service account you created in your azure organization (you could give a try following [this tutorial](https://support.heateor.com/how-to-get-windows-live-client-id-and-client-secret-key/)) and similarly to Google Sign-On, you can enable connection to users from your private organization only, or from any microsoft product world-wide.

Technically, I switched from the library [microsoft-authentication-library-for-js](https://github.com/AzureAD/microsoft-authentication-library-for-js) (formerly passport-azure-ad) to [passport-azure-ad-oauth2](https://github.com/auth0/passport-azure-ad-oauth2). The first one sounds both more maintained and more complete (as it does not support only OAuth2 but also OIDC).
Anyway, I tried to integrate the `microsoft-authentication-library-for-js` lib but without success, the flow does not seem to fit Nhost implementation. Whereas, with the `passport-azure-ad-oauth2` lib I could get authentication working easily with a flow similar to the Apple one.

I would advise these documents:
- [Official presentaiton of Microsoft Azure AD](https://learn.microsoft.com/en-us/azure/active-directory/fundamentals/active-directory-whatis)
- [Official doc of OAuth 2 authentication with Azure AD](https://learn.microsoft.com/en-us/azure/active-directory/fundamentals/auth-oauth2)
- [Difference between the two libs from the author of passport-azure-ad](https://stackoverflow.com/questions/36631122/what-is-the-difference-between-passport-azure-ad-vs-passport-azure-ad-oauth2)

plmercereau (approved):

plmercereau:
Merci Yannick!

PR #251, opened 2022-09-28
fix: drop authenticators inconsistency safely

plmercereau:
drop metadata inconsistency **if and only if** they are related to the authenticators to security key name change

### DIFF
--- a/src/metadata.ts
+++ b/src/metadata.ts
@@ -228,7 +228,29 @@ export const applyMetadata = async (): Promise<void> => {
   logger.info('Applying metadata...');
 
   logger.debug('Reloading metadata...');
-  await reloadMetadata();
+  const { is_consistent, inconsistent_objects } = await reloadMetadata();
+
+  // * Manage metadata inconsistencies
+  if (is_consistent === false) {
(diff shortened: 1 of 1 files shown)


src/metadata.ts
```diff
@@ -228,7 +228,29 @@ export const applyMetadata = async (): Promise<void> => {
...
   logger.debug('Reloading metadata...');
-  await reloadMetadata();
+  const { is_consistent, inconsistent_objects } = await reloadMetadata();
+
+  // * Manage metadata inconsistencies
+  if (is_consistent === false) {
+    // * Drop inconsistencies related to the change from `authenticators` to `security_keys`
+    // * Will drop if and only if the inconsistency is related to this name change
+    if (
+      inconsistent_objects?.length === 2 &&
```
elephant3, line 238: a race condition (parallel deployments of other services, OOM Kill, etc.) may happen between lines 231 and 234, so in my opinion this fix doesn't help in all cases

elephant3: scratch that

elephant3 (approved):

nunopato (approved):

PR #276, opened 2022-11-08
fix(otp): increase OTP secret entropy to 256 bits

### LINKED ISSUE #220: The OTP secrets are generated with only 10 bytes of entropy
Hey,

This is potentially a security vulnerability. I am reporting it here since you neither have a [security policy](https://github.com/nhost/hasura-auth/security) in your repository nor your website uses the https://securitytxt.org/ standard (I would expect https://nhost.io/security.txt to provide contact information). This is also not of a huge impact, but still.

This repo depends on otplib@12.0.1 for OTP secrets generation which uses only 10 bytes of entropy for the OTP secret generation. This may allow someone who knows a few OTP tokens and the time they were generated to brute force the random value used to generate the secret and then predict future generated OTP tokens.

The `hasura-auth` uses the `otplib` for secret generation in here:
https://github.com/nhost/hasura-auth/blob/064b15be7da6eebb9f647d269e2c9cb9a6a795a6/src/routes/mfa/totp-generate.ts#L29-L45

The `totpSecret` is generated from 10 bytes of randomness since this is the default number of bytes the `otplib@12.0.1` uses under the hood. This can be seen [here](https://github.com/yeojz/otplib/blob/v12.0.1/packages/otplib-core/src/authenticator.ts#L262-L267) which uses [this](https://github.com/yeojz/otplib/blob/f418f79dedfc24c3d13aae03bf1a29e64aefedc1/packages/otplib-core/src/authenticator.ts#L166-L174) method under the hood.

By the way, this behavior is also not compliant with some RFCs that recommend to use at least 160 bits of randomness.

Also please note that previous versions of otplib used 20 bytes of randomness instead of 10 for secret generation. I have no idea why this was changed in 12.0.0 to just 10 bytes.

This issue is also reported on the upstream otplib repository here: https://github.com/yeojz/otplib/issues/671 

To fix the issue, you can simply use e.g. `authenticator.generateSecret(32);` which would provide 32 bytes (256 bits) instead of the default (10) bytes of randomness for secret generation.

L3K0V:
Closes #220 

Before submitting this PR:

### Checklist

- [x] No breaking changes
- [x] Tests pass

### Breaking changes

### Tests

### Documentation

### Conventional commits

Versioning and changelog are generated following [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/). Please do your best to follow the convention, but don't worry about it too much. We will be most likely moving away to [changesets](https://github.com/changesets/changesets) soon.

### DIFF
--- a/src/routes/mfa/totp-generate.ts
+++ b/src/routes/mfa/totp-generate.ts
@@ -26,7 +26,7 @@ export const mfatotpGenerateHandler: RequestHandler<
     return sendError(res, 'forbidden-anonymous');
   }
 
-  const totpSecret = authenticator.generateSecret();
+  const totpSecret = authenticator.generateSecret(32);
   const otpAuth = authenticator.keyuri(
     userId,
     ENV.AUTH_MFA_TOTP_ISSUER,


plmercereau:
Thanks for this Asen!

Issue #100, opened 2022-03-01
Activating Twitter sign-in provider causes Hasura Auth to break on startup


elitan:
```
yarn run v1.22.17
$ ts-node-dev -r tsconfig-paths/register --respawn --clear  --unhandled-rejections=strict src/start.ts
[INFO] 13:13:43 ts-node-dev ver. 1.1.8 (using ts-node ver. 9.1.1, typescript ver. 4.5.4)
Error: The OAuth2 adapter used to refresh the token is not configured correctly. Use the setRefreshOAuth2 option to return a OAuth 2.0 adapter.
    at Object.AuthTokenRefresh.use (/Users/eli/code/nhost/hasura-auth/node_modules/passport-oauth2-refresh/lib/refresh.js:86:11)
    at initProvider (/Users/eli/code/nhost/hasura-auth/src/routes/signin/providers/utils.ts:240:13)
    at exports.default (/Users/eli/code/nhost/hasura-auth/src/routes/signin/providers/twitter.ts:9:15)
    at Object.<anonymous> (/Users/eli/code/nhost/hasura-auth/src/routes/signin/providers/index.ts:20:8)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Module._compile (/Users/eli/code/nhost/hasura-auth/node_modules/source-map-support/source-map-support.js:568:25)
    at Module.m._compile (/private/var/folders/kv/klbylzf91lx5ll8mbb2f5hrc0000gn/T/ts-node-dev-hook-766919262572896.js:69:33)
    at Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at require.extensions.<computed> (/private/var/folders/kv/klbylzf91lx5ll8mbb2f5hrc0000gn/T/ts-node-dev-hook-766919262572896.js:71:20)
    at Object.nodeDevHook [as .ts] (/Users/eli/code/nhost/hasura-auth/node_modules/ts-node-dev/lib/hook.js:63:13)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/Users/eli/code/nhost/hasura-auth/src/routes/signin/index.ts:15:1)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
[...]

PR #277, opened 2022-11-08
Grant

plmercereau:
### Improvements
- Providers are defined in [one single place](https://github.com/nhost/hasura-auth/blob/refactor/grant/src/routes/oauth/config.ts) so it is easier to add new ones
- Fix Twitter provider #100
- Fetch display name from the Strava provider
- Fetch the user locale when available (Discord, Google, LinkedIn, WorkOS)
- Fetch the email verification status when available (Apple, BitBucket, Discord, GitHub, Google)
- Fetch avatar url from BitBucket
- The service starts when a provider is incorrectly configured. Instead, the endpoint fails with a standard error. The error is logged (warn level)
- Custom scopes set as environment variables don't replace the scopes that are required by Hasura-auth to function. They are appended instead.
- Profile transformation unit tests

### Deviations

- Don't send the refresh token in the url hash anymore (deprecated since ~6 months)
- Emails that are not explicitly flagged by the provider as `verified` are not set as verified in the database

### To be considered

- [ ]  Implement the disable sign-up feature
- [ ] Ask anyone that contributed in adding an OAuth provider to test this PR
- [ ]  End-to-end tests (requires quite a significant amount of work)


package.json
```diff
@@ -56,27 +51,13 @@
     "js-yaml": "3.13.1",
     "json-to-graphql-query": "^2.2.0",
     "jsonata": "^1.8.5",
+    "jsonwebtoken": "^8.5.1",
```
elitan, line 54: Are we not able to use `jose` instead of importing another jwt package?

src/routes/oauth/config.ts
```diff
@@ -0,0 +1,387 @@
...
+
+  discord: {
+    grant: {
+      client_id: process.env.AUTH_PROVIDER_DISCORD_CLIENT_ID,
+      client_secret: process.env.AUTH_PROVIDER_DISCORD_CLIENT_SECRET,
+      scope: ['identify', 'email'],
+    },
+    profile: ({ profile }) => ({
+      id: profile.id,
+      displayName: `${profile.username}#${profile.discriminator}`,
```
elitan, line 124: I wonder if we should skip `#${profile.discriminator}` here.

If I sign in to an app using Discord I think I would prefer my display name on that app to be `elitan` instead of `elitan#7162`.

src/routes/oauth/index.ts
```diff
@@ -0,0 +1,296 @@
...
+  /**
+   * Use a middleware to keep the session between Oauth requests.
+   * Once the authentication choregraphy is done (either success or failure), the session is destroyed.
+   */
+  .use(
+    session({
+      secret: 'grant',
+      resave: false,
+      saveUninitialized: true,
+      store: new SessionStore(),
```
elitan, line 68: Do we need to use `express-session` here? Just wondering if we could do without that package. The extension of the Store seems pretty trivial.

src/routes/oauth/index.ts
```diff
@@ -0,0 +1,296 @@
...
+   * 5. Generate and return a new refresh token
+   */
+  .use(async ({ session, query }, res) => {
+    const { grant, options, redirectTo = ENV.AUTH_CLIENT_URL } = { ...session };
+    // * Destroy the session as it is only needed for the oauth flow
+    await new Promise((resolve) => {
+      session.destroy(() => {
+        // * Delete the cookie manually
+        // See  https://stackoverflow.com/questions/70101660/how-to-destroy-session
+        res.clearCookie(SESSION_NAME);
```
elitan, line 168: Where is this cookie set?

src/routes/oauth/utils.ts
```diff
@@ -0,0 +1,135 @@
...
+        );
+        if (customScope.length) {
+          // * Adds the scope defined in the env variables to the scope sent as a parameter
+          const initialScope = Array.isArray(grant.scope)
+            ? grant.scope
+            : typeof grant.scope === 'string'
+            ? [grant.scope]
+            : [];
+          // * Merge both sources, and remove duplicates
+          grant.scope = [...new Set([...initialScope, ...customScope])];
```
elitan, line 120: Smart to use a Set.

package.json
plmercereau: unfortunately not: we are loading the Grant config synchronously, and jose is only able to sign JWTs asynchronously. However we could consider replacing `jose` by `jsonwebtoken` - but it needs further thinking and it's better to leave it as it is for now

src/routes/oauth/config.ts
plmercereau: 0 opinion :)
Maybe someone else? @nunopato ?

src/routes/oauth/index.ts
plmercereau: Yes, we still need it so it handles the session during the oauth choreography

src/routes/oauth/index.ts
plmercereau: by `express-session` ;)

src/routes/oauth/config.ts
plmercereau: Although I have a tiny opinion: this information will be a bit complicated for developers to get at a later stage, while it will be easier for them either to modify the display name in the database or to hide the discriminator in their frontend

elitan: Ok, let's keep it as is for now.

elitan: Seems like we had this behaviour before too: https://github.com/oauth-everything/passport-discord/blob/master/src/Strategy.ts#L91

Let's keep it as is.

elitan (approved):

Issue #273, opened 2022-11-03
Query Parameters Being Converted to Lowercase


linden:
we rely on query parameters to determine the source of an function call (i.e. `/api/example?source=Nike`) but somewhere along the line `Nike` gets converted to `nike`, this seems to be a recent change that we first noticed about a week ago.

elitan:
From our Discord discussion, it seems like this issue is happening in production only (not locally), right @linden?

If so, this issue might be some Nhost production gateway issue.

linden:
Hi @elitan, yes this is only happening in production (even with the latest version of the nhost CLI), what can we do to address this if it's a production gateway issue?

PR #279, opened 2022-11-09
Prevent `Joi` from Converting `redirectTo` to Lowercase

linden:
fixes https://github.com/nhost/hasura-auth/issues/273.

### checklist
- [x] no breaking changes.
- [ ] tests pass.
^ couldn't get tests to run in the first place, so unsure if they pass.
- [ ] new features have new tests.
- [ ] documentation is updated.

### DIFF
--- a/src/validation/fields.ts
+++ b/src/validation/fields.ts
@@ -75,7 +75,6 @@ export const redirectTo = ENV.AUTH_CLIENT_URL
       .try(
         ...ENV.AUTH_ACCESS_CONTROL_ALLOWED_REDIRECT_URLS.map((value) =>
           Joi.string()
-            .lowercase()
             .regex(new RegExp('^' + value))
         ),
         Joi.string()
--- a/src/validation/fields.ts
+++ b/src/validation/fields.ts
@@ -75,7 +75,9 @@ export const redirectTo = ENV.AUTH_CLIENT_URL
       .try(
         ...ENV.AUTH_ACCESS_CONTROL_ALLOWED_REDIRECT_URLS.map((value) =>
           Joi.string()
+            .lowercase()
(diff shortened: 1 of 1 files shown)


src/validation/fields.ts
```diff
@@ -75,7 +75,6 @@ export const redirectTo = ENV.AUTH_CLIENT_URL
       .try(
         ...ENV.AUTH_ACCESS_CONTROL_ALLOWED_REDIRECT_URLS.map((value) =>
           Joi.string()
-            .lowercase()
             .regex(new RegExp('^' + value))
```
plmercereau, line 78: Thanks for this!
That would solve the issue, but on the other hand it would make `AUTH_ACCESS_CONTROL_ALLOWED_REDIRECT_URLS` case sensitive.
Maybe something like this?

```suggestion
            .lowercase()
            .regex(new RegExp('^' + value))
            .raw()
```

Ideally we should add tests. If it's too much asking no worries, I'll take over from there and will do my best to do it by Monday

linden: good find, was looking for something like `.raw()` but haven't used `Joi` before so didn't realize that existed. changed in https://github.com/nhost/hasura-auth/pull/279/commits/55ad462398383f3d6d7a889a6d34e4db02684099.

I'd like to contribute more in the future but I don't currently have the time to debug the tests/write them.

plmercereau:
Thanks again @linden !
Just saw your comment about tests in the PR recap.
To run tests:
```
make dev
```

To run tests in watch mode:
```sh
make watch
```

linden:
@plmercereau for some reason `make test` fails locally, likely something with my build environment (considering the github action is working). will debug another time.

PR #282, opened 2022-11-16
feat: 🎸 Allow wildcards and other matching patterns in allowed urls

plmercereau:
Allow wildcards and other [micromatch patterns](https://github.com/micromatch/micromatch#matching-features) in `AUTH_ACCESS_CONTROL_ALLOWED_REDIRECT_URLS`.

To match `https://random-subdomain.vercel.app`:
```
AUTH_ACCESS_CONTROL_ALLOWED_REDIRECT_URLS=https://*.vercel.app
```

It is possible to use other patterns, for instance:

- to allow both http and https:

```
AUTH_ACCESS_CONTROL_ALLOWED_REDIRECT_URLS=http?(s)://website.com
```

- to allow any port:

```
AUTH_ACCESS_CONTROL_ALLOWED_REDIRECT_URLS=http://website.com?(:{1..65536})
```

### Checklist

- [x] No breaking changes
- [x] Tests pass


src/validation/fields.ts
```diff
@@ -69,26 +70,58 @@ export const metadata = Joi.object().default({}).example({
...
-          .regex(new RegExp('^' + ENV.AUTH_CLIENT_URL))
-          // * Get the initial value so we don't change the case
-          .raw()
-      )
-      // ! If validation fails, it will always fall back to `ENV.AUTH_CLIENT_URL` instead of raising an error.
-      .example(`${ENV.AUTH_CLIENT_URL}/catch-redirection`)
-  : Joi.string().uri();
+export const redirectTo = Joi.string()
+  .messages({
+    redirectTo: '{{#label}} is not allowed by the current hasura-auth settings',
```
elitan, line 75:
```suggestion
    redirectTo: '{{#label}} is not allowed as a `redirectTo` parameter.',
```

test/validation.test.ts
```diff
@@ -54,4 +55,96 @@ describe('Unit tests on field validation', () => {
       expect(value).toEqual('+33140506070');
     });
   });
+
+  describe('redirections', () => {
```
elitan, line 59: It seems that we're only checking valid URLs? If that's correct, I think we should also have tests for invalid URLs.

elitan (changes requested):

src/validation/fields.ts
plmercereau: `{{#label}}` is always `"redirectTo"`, so it will read:
```
"redirectTo" is not allowed as a `redirectTo` parameter.
```

test/validation.test.ts
plmercereau: yes indeed

src/validation/fields.ts
elitan: ah ok.

elitan, line 75:
```suggestion
    redirectTo: 'The value of {{#label}} is not allowed.',
```

elitan: How about this?

elitan (changes requested):

elitan (approved):

PR #295, opened 2022-11-29
feat: hash refresh tokens

### LINKED ISSUE #290: Hash refresh tokens
We should hash all refresh tokens (`auth.refresh_tokens`) before we store them in the datbase.

This way, we avoided storing them in plain text, which can lead to bad things in case the refresh tokens are leaked.

plmercereau:
- store hashed refresh tokens with SHA256 in a new separate table `auth.hashed_refresh_tokens`
- keep the previous `auth.refresh_tokens` table to avoid changing the GraphQL schema
- keep the existing refresh tokens to avoid breaking the user's current sessions

Closes #290


src/gql/users.graphql
```diff
@@ -36,14 +36,14 @@ query users($where: users_bool_exp!) {
 }
 
 mutation getUsersByRefreshTokenAndUpdateRefreshTokenExpiresAt(
-  $refreshToken: uuid!
+  $refreshTokens: [String!]!
```
elitan, line 39: This is a breaking change.

migrations/00011_hashed_refresh_tokens.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "auth"."refresh_tokens"
+    ALTER COLUMN "refresh_token" SET DATA TYPE Text;
```
elitan, line 2: This is a breaking change because the data type of the column changes from `uuid` to `text`.

migrations/00011_hashed_refresh_tokens.sql
```diff
@@ -1,3 +1,14 @@
...
+    expires_at timestamp with time zone NOT NULL,
+    user_id uuid NOT NULL
+);
+
+ALTER TABLE auth.hashed_refresh_tokens
+    ADD CONSTRAINT fk_user FOREIGN KEY (user_id) REFERENCES auth.users (id) ON UPDATE CASCADE ON DELETE CASCADE;
+
+COMMENT ON TABLE auth.refresh_tokens IS 'Hashed user refresh tokens. Hasura auth uses them to rotate new access tokens as long as the refresh token is not expired. Don''t modify its structure as Hasura Auth relies on it to function properly.';
+
+COMMENT ON TABLE auth.hashed_refresh_tokens IS 'Deprecated: user refresh tokens. This table is replaced by hashed_refresh_tokens and will be removed in the next major release.';
```
L3K0V, line 13: The comments should we swapped right?

plmercereau: you are right. Correcting

.changeset/four-plums-report.md
```diff
@@ -0,0 +1,10 @@
+---
+'hasura-auth': minor
```
L3K0V, line 2: Should be major a?

plmercereau: no, as we are preventing a breaking change: the existing SQL and GraphQL schemas are not altered, and the ongoing refresh tokens are preserved

.changeset/four-plums-report.md
```diff
@@ -0,0 +1,10 @@
...
+---
+
+Store refresh tokens in SHA256
+
+The existing refresh tokens remains stored in plain text to avoid breaking the ongoing connections
+Hashed refresh tokens are stored in a separate `auth.hashed_refresh_tokens` table. Every new token will be automatically stored in this table.
+
+The use of the `auth.refresh_tokens` table is deprecated, and the table will be removed in the future.
```
elitan, line 10: Should we be more explicit about timeline and future major release here?

Like:

The table `auth.refresh_tokens` is now deprecated and will be removed at [date] in a new major release.

plmercereau: I'm not sure. I would avoid hardcoding a release date, but I would rather put a tentative date in the release notes.
I also think think to set a date is related to our processes rather than this pull request: we can have a discussion about clarifying our release and deprecation policies in parallel to this work

plmercereau:
As it introduces a breaking change, we decided to make the change in two steps:
1. https://github.com/nhost/hasura-auth/pull/311 introduce a `refresh_token_hash` column
2. rename the `refresh_token` column to `id`, and use it as an ID that would not be the refresh token. 

The step 2 will be introduced in the next major release

PR #296, opened 2022-11-30
feat: optionally conceal sensitive error messages

### LINKED ISSUE #294: Prevent guessable user accounts / user enumeration
For data safety reasons, an (at least) an option to prevent user enumeration / guessable user accounts would be highly appreciated. This point was shared by our pen-testing company in a preparation discussion, that this is a common finding. 
I know that in some parts of the world the current behaviour is also desired, our field unfortunately provides some additional struggles.

Some of the current error messages in the authentication part allow it for attackers to try if a user is registered with a specific e-mail. Using / providing options to use neutral error messages.
OWASP describes it as follows:
https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/03-Identity_Management_Testing/04-Testing_for_Account_Enumeration_and_Guessable_User_Account.html

Relevant processes would be (from my current point of view):
- Login
- Signup (email/password) (e.g. the EmailAlreadyInUseError)
- Password reset

An additional thing would be to add the error code to the server errors, to handle the incoming errors by code and not by text-string (which might change). Stumbled across this while migrating from v1 to v2, where those errors texts changed for example. 

Would highly appreciate a chance to toggle / modify the error messages from the backend.

Many thanks!

plmercereau:
Introduce a new `AUTH_CONCEAL_ERRORS` environment variable that conceals error messages to avoid leaking indirect information about users e.g. a user is registered in the application or a given password is invalid.

It is disabled by default.

Closes #294 

### Checklist

- [x] No breaking changes
- [x] Tests pass


elitan (approved):
👏

PR #311, opened 2022-12-15
feat: add a `refresh_token_hash` column

plmercereau:
Introduces a new `refresh_tokens.refresh_token_hash` column.
- the `refresh_tokens.refresh_token` column is deprecated
- the hashed refresh token is a Postgres stored generated column
- the internal GraphQL queries are using the hashed refresh token
- the internal GraphQL mutations are still updating the `refresh_token` column

When introducing the breaking change, we will:
- rename `refresh_tokens.refresh_token` to `refresh_tokens.id`
- use the `id` column as an identifier
- remove the `generated` expression in the `refresh_token_hash` column
- save any new refresh token as a hash


migrations/00011_add_refresh_token_hash_column.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+ALTER TABLE "auth"."refresh_tokens"
+    ADD COLUMN "hashed_refresh_token" VARCHAR(255) GENERATED ALWAYS AS (sha256 (refresh_token::text::bytea)) STORED;
```
elitan, line 3: 1. Why do we generate this value instead of storing the hash? Storing the hash seems more sensible because that's what we'll do in the next major release.
2. I think the column should be named `refresh_token_hash` to be consistent with the naming of `password_hash` (https://github.com/nhost/hasura-auth/blob/main/migrations/00001_create-initial-tables.sql#L59).

elitan (changes requested):

migrations/00011_add_refresh_token_hash_column.sql
plmercereau: 1. Postgres generated columns are actually stored into the database. To "convert" them into standard columns is straight-forward as we only need to remove the generated expression. This approach reduces the burden of having to keep the two columns in sync in hasura-auth during the transition period. We will have to create a new migration regardless, as we will change the primary key name from `refresh_token` to `id`.
2. makes sense, will change this

elitan: 1. sounds good
2. check

.changeset/silver-cars-applaud.md
```diff
@@ -0,0 +1,21 @@
...
+- the hashed refresh token is a Postgres stored generated column
+- the internal GraphQL queries are using the hashed refresh token
+- the internal GraphQL mutations are still updating the `refresh_token` column
+
+When introducing the breaking change, we will:
+
+- rename `refresh_tokens.refresh_token` to `refresh_tokens.id`
+- use the `id` column as an identifier
+- remove the `generated` expression in the `refresh_token_hash` column
+- any new refresh token will then be saved uniquely as SHA256
```
elitan, line 21:
```suggestion
Introduce a new `refresh_tokens.refresh_token_hash` column.

Preparatory work to store refresh tokens as a SHA256 hash.

To avoid a breaking change, the `refresh_tokens.refresh_token` column remains unchanged until the next major release.

- The `refresh_tokens.refresh_token` column is now deprecated.
- The hashed refresh token is a Postgres stored generated column.
- The internal GraphQL queries are using the hashed refresh token.
- The internal GraphQL mutations are still updating the `refresh_token` column.

When introducing the breaking change, we will:

- Rename `refresh_tokens.refresh_token` to `refresh_tokens.id`.
- Use the `id` column as an identifier.
- Remove the `generated` expression in the `refresh_token_hash` column.
- New refresh tokens will then be saved uniquely as SHA256.
```

src/gql/refresh-tokens.graphql
```diff
@@ -4,10 +4,11 @@ mutation insertRefreshToken($refreshToken: authRefreshTokens_insert_input!) {
   }
 }
 
-mutation deleteRefreshToken($refreshToken: uuid!) {
-  deleteAuthRefreshToken(refreshToken: $refreshToken) {
-    refreshToken
-    expiresAt
+mutation deleteRefreshToken($refreshTokenHash: String) {
```
elitan, line 7:
```suggestion
mutation deleteRefreshToken($refreshTokenHash: String!) {
```

elitan: I guess this change requires re-running graphql codegen.

elitan (changes requested):

nunopato (approved):

elitan (approved):

PR #317, opened 2022-12-29
Refactor: Use SQL instead of GraphQL to interact with the DB

### LINKED ISSUE #289: Available to use Hasura experiment Feature for camelCase Support?
Hasura is supporting GraphQL formatted Query since v2.8.0.

Available to use this feature?

plmercereau:
Run DB operations through SQL rather than GraphQL, so the user can pick any [Hasura naming convention configuration](https://hasura.io/docs/latest/schema/postgres/naming-convention/) they want.

As the SQL operations are simpler than the ones generated by Hasura, and as we don't use Hasura anymore to proxy DB operations, we can expect a slight performance increase. This could be benchmarked.

Finally, hasura-auth now only uses the GraphQL API to generate custom claims. We could then limit the hasura-auth metadata sync on startup to the `auth.users` table, and scrap the rest, so the user has even more flexibility over what they want to track, and how. This would also fasten the start of the hasura-auth service.

Closes #289

### Checklist

- [x] No breaking changes
- [x] Tests pass
- [x] ~~New features have new tests~~


src/routes/signup/webauthn/verify.ts
```diff
@@ -65,27 +65,31 @@ export const signInVerifyWebauthnHandler: RequestHandler<
     });
   }
 
-  const {
-    users: [user],
-  } = await gqlSdk.users({ where: { currentChallenge: { _eq: challenge } } });
-
+  const user = await pgClient.getUserByChallence(challenge);
```
nunopato, line 68:
```suggestion
  const user = await pgClient.getUserByChallenge(challenge);
```

src/utils/postgres-client/index.ts
```diff
@@ -0,0 +1,379 @@
...
+      rows: [user],
+    } = await client.query<SqlUser>(
+      createUserQueryWhere(`u.ticket = $1 and u.ticket_expires_at > NOW()`),
+      [ticket]
+    );
+    client.release();
+    return cameliseUser(user);
+  },
+
+  getUserByChallence: async (challenge: string) => {
```
nunopato, line 296:
```suggestion
  getUserByChallenge: async (challenge: string) => {
```

src/utils/postgres-client/index.ts
```diff
@@ -0,0 +1,379 @@
...
+  cameliseUser,
+  createUserQueryByColumn,
+  createUserQueryWhere,
+  getUserById,
+  insertUserRoles,
+  snakeiseUser,
+} from './utils';
+
+const pool = new Pool({
+  max: 20,
```
nunopato, line 20: I would suggest having this number configurable and default to 50 ([same as Hasura's](https://hasura.io/docs/latest/api-reference/syntax-defs/#pgpoolsettings)).

src/utils/postgres-client/index.ts
```diff
@@ -0,0 +1,379 @@
...
+  insertUser: async (user: Partial<User>) => {
+    const client = await pool.connect();
+    const transformedUser = snakeiseUser(user) as Partial<SqlUser>;
+    const { roles, ...rest } = transformedUser;
+    const columns = Object.keys(rest);
+    const values = Object.values(rest);
+    const {
+      rows: [insertedUser],
+    } = await client.query<SqlUser>(
+      `INSERT INTO "auth"."users" (${columns.join(',')}) VALUES(${[
```
nunopato, line 347: this is necessary because the user columns being filled will depend on the sign-up method used, correct?

plmercereau: Exactly

src/utils/postgres-client/index.ts
plmercereau: I changed to 50, but I would rather not add a configuration until we need to use different values in our cloud, or someone asks for it (we already defined more than 100 environment variables, that's a crowd)

src/routes/signin/passwordless/email.ts
```diff
@@ -117,7 +114,8 @@ export const signInPasswordlessEmailHandler: RequestHandler<
       link,
       displayName: user.displayName,
       email,
-      newEmail: user.newEmail,
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      newEmail: user.newEmail!,
```
elitan, line 118: How about we check `user.newEmail` in an if statement before and throw if it's unavailable? 

Feels cleaner than adding an eslint comment.

src/routes/user/email/send-verification-email.ts
```diff
@@ -93,7 +93,8 @@ export const userEmailSendVerificationEmailHandler: RequestHandler<
       link,
       displayName: user.displayName,
       email: user.email,
-      newEmail: user.newEmail,
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      newEmail: user.newEmail!,
```
elitan, line 97: Same, if-statement before instead of eslint comment

src/utils/postgres-client/utils.ts
```diff
@@ -0,0 +1,163 @@
...
+    otpHash: otp_hash,
+    otpMethodLastUsed: otp_method_last_used,
+    webauthnCurrentChallenge: webauthn_current_challenge,
+    ticketExpiresAt: ticket_expires_at,
+    otpHashExpiresAt: otp_hash_expires_at,
+    lastSeen: last_seen,
+  };
+};
+
+export const snakeiseUser = (
```
elitan, line 65: What does "snakeise" mean? :>

src/utils/postgres-client/index.ts
```diff
@@ -0,0 +1,379 @@
...
+  },
+
+  getUserByRefreshToken: async (refreshToken: string) => {
+    const client = await pool.connect();
+    const { rows } = await client.query<{ id: string }>(
+      `SELECT u.id FROM "auth"."refresh_tokens" rt, "auth"."users" u
+        WHERE rt.user_id = u.id
+          AND rt.refresh_token_hash = $1
+          AND u.disabled = false
+          AND rt.expires_at > NOW();`,
```
elitan, line 234: I'd suggest re-writing to this to increase readability:

```SQL
SELECT u.id 
FROM auth.refresh_tokens AS rt
JOIN auth.users AS u ON rt.user_id = u.id
WHERE rt.refresh_token_hash = $1
  AND u.disabled = false
  AND rt.expires_at > NOW();
```

elitan (changes requested):

elitan (approved):

src/routes/signin/passwordless/email.ts
plmercereau: This was due to incorrect types, as `newEmail` is only mandatory when changing the email. II improved the template typings.

src/utils/postgres-client/index.ts
nunopato: @plmercereau sounds good :+1:

nunopato (approved):

PR #320, opened 2023-01-05
Set custom claims to empty array when value is null | undefined

### LINKED ISSUE #319: Set some custom claims to empty array when value is null | undefined
We came accros an issue with custom claims **when they are expected to be an array**. 

Let's say you define:

- a custom claim looking like this: `"projectsLeaded":"projectsLeaded[].projectId"`.
- some permissions in Hasura in the form of `project.id IN X-Hasura-projectsLeaded`.

With the current implementation of Hasura Auth, if the user doesn't lead any project, the `projectsLeaded` custom claim won't be defined at all, and Hasura will throw an error saying `X-Hasura-projectsLeaded` is required to make the call.
We would rather expect `X-Hasura-projectsLeaded` to be set as an empty array, so that the permissions work as expected in Hasura.

I did a change on our fork to fix this: https://github.com/onlydustxyz/hasura-auth/commit/5a7029a246e7a03268a89bc95a9454da9e7cd3ab

As you can see, I considered that when the custom claim's path contains `[]`, then it is expected to be an array, and consequently, I set its value to an empty array when it is undefined | null.

I think it would make sense to integrate this change to Hasura Auth officially.
What do you think? 

I can of course make a proper PR in case you agree on the principle.

ofux:
Closes #319

### Checklist

- [ ] No breaking changes
- [x] Tests pass

### Breaking changes

Until now, custom claims containing `[]` (ie. expected to be an array) were not defined at all if the array was empty.
With this PR, such claims will now be equal to an empty array (ie. `{}`).

This is a breaking change, however, I feel like this behavior is more intuitive and probably the one expected by users when defining such claims (at least, it is my case).

Please see #319 for more details.

### Documentation

I'm not sure the documentation needs to be updated about this. If it does, please let me know and I will do it.

### DIFF
--- a/.changeset/curvy-rice-sin.md
+++ b/.changeset/curvy-rice-sin.md
@@ -1,0 +1,8 @@
+---
+'hasura-auth': patch
+---
+
+A custom claim that is expected to be an array (ie. contains "[]" in its path) will be set to an empty array - instead of being undefined - when its query returns no value.
+
+This allows permissions of the form "something IN X-Hasura-myCustomClaimArray" to work as intended
+when the array is empty.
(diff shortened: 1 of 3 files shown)


ofux:
Also, please let me know if I should create a changeset for this PR.

ofux:
I just added the changeset

PR #323, opened 2023-01-09
fix: user roles are not properly aggregated

ofux:
Before submitting this PR:

### Checklist

- [x] No breaking changes
- [x] Tests pass
- [ ] ~~New features have new tests
- [ ] ~~Documentation is updated

### The bug

Only one role is returned by the SQL query that fetches user information.

This was not visible from the unit tests because the default role is automatically added afterward (even if not returned by the SQL query). In the end,
tests were working because of luck. I updated a test to make sure it fails before fixing the bug.

### Tests

- Note: changing `AUTH_USER_DEFAULT_ALLOWED_ROLES` with `/change-env` has no effect as the `Joi` constraint is built only once when the application starts.
- The test 'should sign in user and return valid tokens' has been modified to check the very specific bug that is fixed by this PR.
- Other tests have been fixed and updated because they were **not** testing what they intend to test.

### DIFF
--- a/src/utils/postgres-client/utils.ts
+++ b/src/utils/postgres-client/utils.ts
@@ -137,7 +137,7 @@ export const createUserQueryWhere = (where: string) =>
             SELECT user_id, coalesce(json_agg(role), '[]') AS roles
             FROM "auth"."user_roles" r
             WHERE r.user_id = u.id
-            GROUP BY user_id, role
+            GROUP BY user_id
         ) r ON r.user_id = id WHERE ${where};`;
 
 export const createUserQueryByColumn = (column: string) =>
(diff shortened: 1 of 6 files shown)


ofux:
Hi @plmercereau 
I made this fix after discovering this critical bug in our own end2end test suite ;)

.changeset/angry-teachers-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+'hasura-auth': patch
+---
+
+Fix user roles aggregation. Only one role was returned by the SQL query.
```
plmercereau, line 5: there's no need to add a changeset as this bug was introduced by code that is not yet published. You can remove the file

ofux: done

plmercereau (approved):
many thanks!

PR #324, opened 2023-01-10
fix: error missing FROM-clause entry for table "up"

ofux:
Before submitting this PR:

### Checklist

- [x] No breaking changes
- [x] Tests pass

Fixes a bug that makes the application crash when a user signs up for the first time with (at least) github (but I think it may impacts any provider).

The error was: `missing FROM-clause entry for table "up"`.


ofux:
I discovered this bug while testing our app together with the latest Hasura-Auth

plmercereau (approved):

PR #325, opened 2023-01-10
Multiple fixes in Postgres client

ofux:
- fix: error Cannot read properties of undefined (reading 'id')
- fix: error INSERT has more expressions than target columns
- fix: wrong call to insertUserProviderToUser

Before submitting this PR:

### Checklist

- [x] No breaking changes
- [x] Tests pass

Those bugs were spoted while testing Hasura-Auth with Github provider.
They occured when a brand new user (ie. a user who never interacted with the app before) was trying to signup.

### DIFF
--- a/src/routes/oauth/index.ts
+++ b/src/routes/oauth/index.ts
@@ -246,8 +246,8 @@ export const oauthProviders = Router()
           userId: user.id,
           providerId: provider,
           providerUserId,
-          accessToken,
           refreshToken,
+          accessToken,
         });
 
         if (!result) {
@@ -263,8 +263,8 @@ export const oauthProviders = Router()
           userId: id,
           providerId: provider,
           providerUserId,
-          accessToken,
           refreshToken,
+          accessToken,
         });
       }
     }
--- a/src/utils/postgres-client/index.ts
+++ b/src/utils/postgres-client/index.ts
@@ -319,7 +319,7 @@ export const pgClient = {
       user = await getUserById(client, userId);
     }
     client.release();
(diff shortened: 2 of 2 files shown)


plmercereau:
We run the tests defined in this repo, and the e2e tests from the nhost/nhost, but unfortunately they don’t cover oauth nor webauthn.
Oauth is extremely hard to simulate in CI as most providers prevent signin in from puppeteer and the like.
If you have an idea on how to tackle it I’m all ears :)

ofux:
> We run the tests defined in this repo, and the e2e tests from the nhost/nhost, but unfortunately they don’t cover oauth nor webauthn. Oauth is extremely hard to simulate in CI as most providers prevent signin in from puppeteer and the like. If you have an idea on how to tackle it I’m all ears :)

I see. I'll try to come up with something on my own CI as I should ultimately test a real signup and let you know if I find a way to do this properly.

plmercereau:
that would be awesome! Maybe the solution lies into running tests on one single provider, and in identifying one provider that did not set security measures.
Or maybe to find some kind of a mock oauth provider that we could add to the list for testing purpose

ofux:
Another possibility might be to actually embrace the security mesure. Github sends an email to check the device and asks for a code. One could [use gmail API to read last received email](https://stateful.com/blog/gmail-api-node-tutorial) and actually enter the verification code in github login workflow...
This might work, but seems like a lot of work to put in place... might be worth it though...

plmercereau:
Tried this already, GitHub, Google etc actually detect the authentication comes from a bot, and block the attempt. It becomes worst when ran from a GH runner

PR #365, opened 2023-04-18
feat: Personal Access Tokens

### LINKED ISSUE #364: support for PAT in hasura-auth
# Motivation

Today when a user needs to interact with a hasura-auth/hasura backed service outside of a frontend application they have two options:

1. Share the admin secret
2. Implement something customized (like we did ourselves for the cli)

# Goal

The goal is to be able to generate tokens to authenticate a backend service or tool as a user. For instance, a `cli` tool, or some backend service that needs access to hasura.

# Potential implementation

A potential implementation could to simply rely on the existing refresh-tokens functionality. In this scenario, a tool like the cli could:

1. Authenticate the user as normally done
2. Have the user request a PAT
3. Store it

At this point the CLI could just reuse that refresh token and act on behalf of the user.

A similar workflow could be used to authenticate backend services:

1. An admin could create a "service account"
2. Give it a role that has permissions to do what the backend service needs to do
3. Create a PAT/refresh_token
4. Configure it in the backend service

## Requirements

For this to work I guess we need:

1. A mechanism to generate refresh tokens with a longer timespan, i.e., 1 year, 2 years, 99 years....
2. A mechanism to list them and invalidate them

## Security implications

The benefit of this mechanism is that the PAT needs to be authenticated every time to generate an access token so it can be revoked anytime.

szilarddoro:
Resolves #364

### DIFF
--- a/migrations/00013_add_refresh_token_type.sql
+++ b/migrations/00013_add_refresh_token_type.sql
@@ -1,0 +1,5 @@
+BEGIN;
+CREATE TYPE refresh_token_type AS ENUM ('regular', 'pat');
+ALTER TABLE auth.refresh_tokens
+ADD COLUMN type refresh_token_type NOT NULL DEFAULT 'regular';
+END;
--- a/src/routes/pat/pat.ts
+++ b/src/routes/pat/pat.ts
@@ -31,7 +31,10 @@ export const createPATHandler: RequestHandler<
   // Note: Joi wouldn't work here because we need to compare the date to the
   // date of the request, not the date when the schema was created
   // 7 days
-  if (new Date(expiresAt).getTime() < Date.now() + 7 * 24 * 60 * 60 * 1000) {
+  if (
+    new Date(expiresAt).setHours(0, 0, 0, 0) <
+    new Date().setHours(0, 0, 0, 0) + 7 * 24 * 60 * 60 * 1000
+  ) {
     return sendError(res, 'invalid-expiry-date', {
       customMessage: 'The expiry date must be at least 7 days from now',
     });
--- a/src/routes/pat/pat.ts
+++ b/src/routes/pat/pat.ts
@@ -12,7 +12,7 @@ export const createPATSchema = Joi.object({
 export const createPATHandler: RequestHandler<
   {},
   {},
-  { refreshToken: string; metadata: object; expiresAt: Date }
+  { metadata: object; expiresAt: Date }
 > = async (req, res) => {
   if (!req.auth) {
     return sendError(res, 'unauthenticated-user');
--- a/src/routes/pat/pat.ts
+++ b/src/routes/pat/pat.ts
@@ -1,0 +1,57 @@
+import { sendError } from '@/errors';
+import { getUser, gqlSdk } from '@/utils';
+import { RequestHandler } from 'express';
+import Joi from 'joi';
+import { v4 as uuidv4 } from 'uuid';
+
+export const createPATSchema = Joi.object({
+  expiresAt: Joi.date().required(),
+  metadata: Joi.object(),
+}).meta({ className: 'CreatePATSchema' });
+
+export const createPATHandler: RequestHandler<
+  {},
+  {},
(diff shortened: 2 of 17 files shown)


migrations/00013_add_refresh_token_type.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
```
dbarrosop, line 1: does this work with hasura?

dbarrosop: it actually works, this must be new. They used to suggest using a different table for this. Nice!

src/routes/pat/pat.ts
```diff
@@ -0,0 +1,60 @@
+import { sendError } from '@/errors';
+import { getUser, gqlSdk } from '@/utils';
+import { RequestHandler } from 'express';
+import Joi from 'joi';
+import { v4 as uuidv4 } from 'uuid';
+
+export const createPATSchema = Joi.object({
+  expiresAt: Joi.date().required(),
+  metadata: Joi.object(),
```
dbarrosop, line 9: what is this metadata?

dbarrosop: ignore me

src/routes/pat/pat.ts
```diff
@@ -0,0 +1,60 @@
...
+}).meta({ className: 'CreatePATSchema' });
+
+export const createPATHandler: RequestHandler<
+  {},
+  {},
+  { refreshToken: string; metadata: object; expiresAt: Date }
```
dbarrosop, line 15: shouldn't `refreshToken` be an `accessToken`? In which case you'd just validate the signature and create a PAT for the user specified in the JWT

dbarrosop:
The workflow looks good, should we provide a mechanism for:

- getting PATs
- revoking/deleting PATs

Or is the expectation people will use graphql queries for that? I think that's it, right?

dbarrosop (approved):

src/routes/pat/pat.ts
```diff
@@ -0,0 +1,60 @@
...
+  if (!req.auth) {
+    return sendError(res, 'unauthenticated-user');
+  }
+
+  const { userId } = req.auth as RequestAuth;
+
+  const user = await getUser({ userId });
+
+  if (!user) {
+    return sendError(res, 'invalid-refresh-token');
```
nunopato, line 26: did you mean user instead of refresh token here?

src/routes/pat/pat.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  const { metadata, expiresAt } = req.body;
+
+  // Note: Joi wouldn't work here because we need to compare the date to the
+  // date of the request, not the date when the schema was created
+  // 7 days
```
nunopato, line 33: out of curiosity, what is the reasoning for 7 days?

szilarddoro: I followed GitHub's approach in this. The minimum amount of expiration for their legacy tokens is 7 days. We can change it to anything.

src/routes/pat/pat.ts
szilarddoro: Good catch. Will fix this.

src/routes/signin/pat.ts
```diff
@@ -0,0 +1,37 @@
...
+export const signInPATSchema = Joi.object({
+  personalAccessToken,
+}).meta({ className: 'SignInPATSchema' });
+
+export const signInPATHandler: RequestHandler<
+  {},
+  {},
+  { personalAccessToken: string }
+> = async (req, res) => {
+  const user = await getUserByRefreshToken(req.body.personalAccessToken);
```
dbarrosop, line 16: should we also validate the type of the refresh token is `PAT`?

szilarddoro: Yeah, good point. I'll create a separate GraphQL query for this one.

src/routes/pat/pat.ts
nunopato: :+1:

PR #372, opened 2023-04-26
chore: improve log verbosity

szilarddoro:


nunopato (approved):

dbarrosop:
can we also log why the 5xx and 4xx errors? Right now it only logs the request, but not the reason for the issue

PR #379, opened 2023-05-05
chore(logs): add env var to control visibility in logs

szilarddoro:


dbarrosop:
Should we update [the environment variables docs](https://github.com/nhost/hasura-auth/blob/main/docs/environment-variables.md)? Make emphasis this isn't recommended in production as it will log user credentials and other sensitive information.

dbarrosop (approved):

Issue #221, opened 2022-07-28
Database migrations deadlock when starting with a fresh DB.


Tiim:
I'm trying to use hasura-auth as a selfhosted service with docker. In development this works really nice. Now I want to deploy my application and I got stuck on the following:

- hasura-auth does not run the migrations because hasura is not started up. (Hasura is not ready. Retry in 5 seconds.)
- hasura tries to apply the migrations before starting up (dockerfile: hasura/graphql-engine:v2.8.4.cli-migrations-v3), but fails because some tables have a foreign key that link to the users table which does not exist yet.

Now both hasura and hasura-auth are in a deadlock where both are waiting for each other to apply the migrations.

It would be nice if hasura-auth could apply the database migrations even if hasura has not started yet.

Bernix01:
This issue is still valida (? closed and marked stale...

dbarrosop:
Following the conversation here: https://github.com/nhost/hasura-auth/pull/389#issuecomment-1773915106

PR #389, opened 2023-05-19
Apply migrations before waiting for Hasura.

2start:
**Problem:** The current docs suggest to extend the user schema by creating a new table in schema public with Hasura and link via a FK to the auth.user schema ([See Docs](https://github.com/nhost/hasura-auth/blob/main/docs/recipes/extending-user-schema.md)). This does not work because Nhost and Hasura may have a circular dependency on each other if you create a new environment from scratch:

Hasura migrations depend may depend on auth.user. 
Auth Service depends on Hasura to be healthy first (i.e. having migrations applied) because the order of execution is as follows:

  await waitForHasura();
  await applyMigrations();
  await applyMetadata();
  
**Solution:** 

Just change the order of execution to: 

  await applyMigrations();
  await waitForHasura();
  await applyMetadata();
  
This works because the migrations have no dependency on Hasura at all.
It fixed our problem with this circular dependency.

Also curious how this was not noticed until now. It's basically not possible to build a new environment from scratch with existing metadata/migrations when referencing auth.user in the Hasura migrations.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -19,10 +19,10 @@ export const start = async () => {
   const { applyMigrations } = await import('./migrations');
   const { applyMetadata } = await import('./metadata');
 
-  // wait for hasura to be ready
-  await waitForHasura();
   // apply migrations and metadata
   await applyMigrations();
+  // wait for hasura to be ready
+  await waitForHasura();
   await applyMetadata();
   // }
 


2start:
@plmercereau
@elitan

Could you please take a look at this. I put in the effort to write up the problem as concisely as possible. Would be happy to be able to merge this.

Bernix01:
any update on this?

dbarrosop:
Sorry, we missed this PR somehow. I don't think this is a problem in hasura-auth. From the looks of it, it looks like:

1. You have a project that references the auth.users table, which is fine
2. When you instantiate that project, you are trying to run the migrations for that project before starting hasura-auth for the first time so the auth schema doesn't exist yet.

This looks to me like a problem on the way operations are being performed. Your application depends on hasura-auth but you are trying to instantiate it before hasura-auth.

> Also curious how this was not noticed until now

I am guessing you must be self-hosting because this issue doesn't occur with our cli or cloud offering as things are instantiated in the correct order.

xmlking:
I also experienced same problem. I have to run them manically after servers start. This really help automating this 
https://github.com/xmlking/spectacular/blob/301f260994f133ab8175857c5af9ff6635f70d2e/nhost/migrations/default/1684206403640_init/up.sql#L242

xmlking:
Thanks I will try adding my migrations container depend on auth and hasura and share the results

vanHekthor:
@xmlking were you able to resolve your issue? For me, this really does not seem like a deployment issue - at least in the way described here.
I feel like this PR was prematurely closed.

Here is my understanding of the issue:
The [docs](https://github.com/nhost/hasura-auth/blob/main/docs/recipes/extending-user-schema.md) clearly suggest referencing the auth.users table.
@dbarrosop You are also saying that is fine.

> You have a project that references the auth.users table, which is fine

This means there has to be a migration which creates the table referencing auth.users in the public schema.

For this migration to be applied, the auth schema needs to be present. Therefore hasura-auth has to be instantiated.

But hasura-auth itself waits for hasura, the graphql-engine to be ready (see [here](https://github.com/nhost/hasura-auth/blob/a6d3f8e9e5555a8e35fc7d411f65ab0a79a9afb0/src/server.ts#L22-L27)).

Hasura is never ready, because the aforementioned migration can't be applied.

And there is our deadlock:
Hasura can't apply all migrations because the auth schema is missing
-> the auth schema is missing because hasura-auth is not instantiated
-> hasura-auth is not instantiated because it waits for hasura to be ready
-> hasura is not ready because it can't apply all its migrations without the auth schema
-> the auth schema ...

This was already addressed in this [issue](https://github.com/nhost/hasura-auth/issues/221#issue-1321140127), where Tiim suggests:

> It would be nice if hasura-auth could apply the database migrations even if hasura has not started yet.

This PR by @2start proposes exactly this change and it should work because as 2Start says the hasura-auth migrations have no dependency to hasura at all.

xmlking:
> @xmlking were you able to resolve your issue? For me, this really does not seem like a deployment issue - at least in the way described here. I feel like this PR was prematurely closed.
> 
> Here is my understanding of the issue: The [docs](https://github.com/nhost/hasura-auth/blob/main/docs/recipes/extending-user-schema.md) clearly suggest referencing the auth.users table. @dbarrosop You are also saying that is fine.
> 
> > You have a project that references the auth.users table, which is fine
> 
> This means there has to be a migration which creates the table referencing auth.users in the public schema.
> 
> For this migration to be applied, the auth schema needs to be present. Therefore hasura-auth has to be instantiated.
> 
> But hasura-auth itself waits for hasura, the graphql-engine to be ready (see [here](https://github.com/nhost/hasura-auth/blob/a6d3f8e9e5555a8e35fc7d411f65ab0a79a9afb0/src/server.ts#L22-L27)).
> 
> Hasura is never ready, because the aforementioned migration can't be applied.
> 
> And there is our deadlock: Hasura can't apply all migrations because the auth schema is missing -> the auth schema is missing because hasura-auth is not instantiated -> hasura-auth is not instantiated because it waits for hasura to be ready -> hasura is not ready because it can't apply all its migrations without the auth schema -> the auth schema ...
> 
> This was already addressed in this [issue](https://github.com/nhost/hasura-auth/issues/221#issue-1321140127), where Tiim suggests:
> 
> > It would be nice if hasura-auth could apply the database migrations even if hasura has not started yet.
> 
> This PR by @2start proposes exactly this change and it should work because as 2Start says the hasura-auth migrations have no dependency to hasura at all.

We have to orchestrate the start order of 1. hsura-graphql 2. Hasura-auth 3. Hasura-migration services

PR #405, opened 2023-07-11
fix #404 Azure not storing users email

### LINKED ISSUE #404: Signing up with Azure AD does not set the users email
When using AzureAD for signup/ signin, the users email is not set in the database.
[This](https://github.com/nhost/hasura-auth/blob/87fee13f4e7b3ea33263323a27ecd3ac556f6eeb/src/routes/oauth/config.ts#L69C1-L88) part of the azuread config is the falsey part. The payload field containing the email is named "upn", in the config it's [email](https://github.com/nhost/hasura-auth/blob/87fee13f4e7b3ea33263323a27ecd3ac556f6eeb/src/routes/oauth/config.ts#L85)

```  
azuread: {
    grant: {
      oauth: 2,
      scope_delimiter: ' ',
      client_id: process.env.AUTH_PROVIDER_AZUREAD_CLIENT_ID,
      client_secret: process.env.AUTH_PROVIDER_AZUREAD_CLIENT_SECRET,
      authorize_url: `${azureBaseUrl}/[subdomain]/oauth2/authorize`,
      access_url: `${azureBaseUrl}/[subdomain]/oauth2/token`,
      profile_url: `${azureBaseUrl}/[subdomain]/openid/userinfo`,
      subdomain: process.env.AUTH_PROVIDER_AZUREAD_TENANT || 'common',
    },
    profile: ({ jwt }) => {
      const payload = jwt?.id_token?.payload;
      return {
        id: payload.oid,
        displayName: payload.name,
        email: payload.email,
      };
    },
  },
```

Here is a privatized OAuth response from AzureAD:

```
{
  "id_token": "...",
  "access_token": "...",
  "refresh_token": "...",
  "jwt": {
    "id_token": {
      "header": {
        "typ": "JWT",
        "alg": "none"
      },
      "payload": {
        "aud": "...",
        "iss": "https://sts.windows.net/.../",
        "iat": 1689073876,
        "nbf": 1689073876,
        "exp": 1689077776,
        "amr": [
          "pwd",
          "mfa"
        ],
        "family_name": "Doe",
        "given_name": "John",
        "ipaddr": "12.345.78.912",
        "name": "John Doe",
        "oid": "...",
        "onprem_sid": "...",
        "rh": "...",
        "sub": "...",
        "tid": "...",
        "unique_name": "email@email.com",
        "upn": "email@email.com",
        "ver": "1.0"
      },
      "signature": ""
    }
  },
  "profile": {
    "error": {}
  }
}
```

I hope this can be fixed quickly, with the provided information.

hk-meko:
Fixes #404

### DIFF
--- a/src/routes/oauth/config.ts
+++ b/src/routes/oauth/config.ts
@@ -82,7 +82,7 @@ export const PROVIDERS_CONFIG: Record<
       return {
         id: payload.oid,
         displayName: payload.name,
-        email: payload.email,
+        email: payload.upn,
       };
     },
   },


dbarrosop:
Thanks for the PR, my only concern is that the [docs](https://learn.microsoft.com/en-us/azure/active-directory/develop/access-token-claims-reference#v10-basic-claims) say this:

> The username of the user. May be a phone number, email address, or unformatted string. Only use for display purposes and providing username hints in reauthentication scenarios.

Not sure if we should validate the value and return an error if it's not an email.

hk-meko:
You are right, but if I see it correctly there is already a validation [set up](https://github.com/nhost/hasura-auth/blob/87fee13f4e7b3ea33263323a27ecd3ac556f6eeb/src/routes/oauth/utils.ts#L38).

dbarrosop (approved):

dbarrosop:
Reading more about this, I am not an Azure AD expert but I am thinking this may not be correct. The UPN doesn't seem to be the email and googling around a bit you should be able to ask for the email claim:

https://stackoverflow.com/questions/65128383/cant-get-email-claim-in-access-token-in-azure-ad

dbarrosop:
Thanks! closing PR and issue, feel free to re-open if needed.

hk-meko:
Could be considered, adding the info to the docu in the future.

xmlking:
@dbarrosop 

stackoverflow https://stackoverflow.com/a/49679604 suggest, **email** can be mapped to different  fields based on what azure subscription company has:  
> For personal accounts, the email address is returned in an **email** field like one would expect.
> For company accounts, the email address is returned in a **preferred_username** field.
 
For my case Azure AD  is sending email at two places:   1. preferred_username 2. upn

image

can you provide options for developers to map email address via configuration ? or may be fallback with:

 
```js
return {
  id: payload.oid,
  displayName: payload.name,
  email: payload.email ?? payload.preferred_username ?? payload.upn
};
```

xmlking:
b.t.w I also tried adding Permissions suggested above, but email claim is not comming. 
 
image

image

Issue #406, opened 2023-07-11
Auth_refresh_token permissions for new role are deleted at nhost up


juicygoose:
Hi,

I have created a new role called `readOnlyClient` and given the following permissions to this role:
```
select_permissions:
  - role: readOnlyClient
    permission:
      columns:
        - id
        - metadata
        - user_id
      filter: {}
```
in the file `auth_refresh_tokens.yaml`

Every time I boot my application, the permission is erased and I have the following git diff :
![image](https://github.com/nhost/hasura-auth/assets/15606243/3d1d70a4-ef32-4680-a637-5bfc736351d2)

Would it be possible to have a look as it is impacting some of the features used in my application ?

Many thanks in advance,
Alex

PR #407, opened 2023-07-11
bugfix(hasura-auth) keep select and delete permissions already set in metadata for auth tables instead of overriding them

juicygoose:
The documentation already mentions that it is possible to set permissions on auth tables.

Fixes https://github.com/nhost/hasura-auth/issues/406

### DIFF
--- a/.changeset/flat-parrots-remain.md
+++ b/.changeset/flat-parrots-remain.md
@@ -1,0 +1,5 @@
+---
+'hasura-auth': patch
+---
+
+fix: merge new select/delete permissions with the existing ones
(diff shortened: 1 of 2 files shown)


onehassan (commented):
Thank you reporting this and taking the time to fix this 🚀

onehassan:
@juicygoose it would great also if you could a changeset as mentioned [here](https://github.com/nhost/hasura-auth/blob/main/DEVELOPERS.md#changesets)

juicygoose:
> @juicygoose it would be great also if you could add a changeset as mentioned [here](https://github.com/nhost/hasura-auth/blob/main/DEVELOPERS.md#changesets)

Would it be possible for you to do it ? I have crashed linked to pnpm and node versions when trying to do it...

PR #425, opened 2023-10-26
fix: incorrect assignment of `displayName` for Apple OAuth

totzk9:
https://developer.apple.com/documentation/sign_in_with_apple/request_an_authorization_to_the_sign_in_with_apple_server#4066168
or
https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/incorporating_sign_in_with_apple_into_other_platforms#3332115

### DIFF
--- a/src/routes/oauth/config.ts
+++ b/src/routes/oauth/config.ts
@@ -53,9 +53,10 @@ export const PROVIDERS_CONFIG: Record<
     },
     profile: ({ jwt }) => {
       const payload = jwt?.id_token?.payload;
+      const user = jwt?.user;
       // * See https://developer.apple.com/forums/thread/118209
-      const displayName = payload?.name
-        ? `${payload.name.firstName} ${payload.name.lastName}`
+      const displayName = user?.name
+        ? `${user.name.firstName} ${user.name.lastName}`
         : payload.email;
       return {
         id: payload.sub,
--- a/src/routes/oauth/config.ts
+++ b/src/routes/oauth/config.ts
@@ -65,7 +65,9 @@ export const PROVIDERS_CONFIG: Record<
             ? `${userProfile.name.firstName} ${userProfile.name.lastName}`
             : displayName;
         } catch (error) {
-          logger.warn(error);
+          logger.warn(
+            `Problem trying to parse user data from apple's response: ${error}`
+          );
 
           // use the user's email as fallback
           displayName = payload.email;
--- a/src/routes/oauth/config.ts
+++ b/src/routes/oauth/config.ts
@@ -53,17 +53,14 @@ export const PROVIDERS_CONFIG: Record<
     },
     profile: ({ jwt, profile }) => {
       const payload = jwt?.id_token?.payload;
-      let user;
-      try {
-        user = JSON.parse(profile);
-      } catch {
-        user = profile;
-      }
 
-      // * See https://developer.apple.com/forums/thread/118209
-      const displayName = user?.name
-        ? `${user.name.firstName} ${user.name.lastName}`
-        : payload.email;
+      const isProfileError = typeof profile === 'object' && profile.error;
+      const userProfile = !isProfileError ? JSON.parse(profile) : null;
+      const displayName =
+        isProfileError && !userProfile
+          ? payload.email
+          : `${userProfile.name.firstName} ${userProfile.name.lastName}`;
+
       return {
         id: payload.sub,
         displayName,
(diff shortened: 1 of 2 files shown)


totzk9:
Hi, I haven't tested this yet and I don't know how to setup hasura-auth to my local.

Based from docs, I believe this should be the correct response

dbarrosop:
Thanks a lot for the PR.

To test locally you can run:

```
docker build -t nhost/hasura-auth:local .
```

Then, if you have an nhost project using the cli you can temporarily set the auth version to `local` to use this image.

When you manage to test it and verify it works send us a few screenshots capturing the result (or even a loom video) and we will merge. If you need help with the test don't hesitate to ping me on discord.

totzk9:
@dbarrosop Will do. thanks!

totzk9:
Here are my local tests with the following videos.

1. Testing the current v0.21.2. The current implementation doesn't properly fetch the user's info from Apple

https://github.com/nhost/hasura-auth/assets/24382050/42dfbae7-2b0a-4e58-b5fd-83defa86e7f9

2. I've added logs to see and compare the json data taken from `jwt` and `profile` params.

https://github.com/nhost/hasura-auth/assets/24382050/edd7e73c-dd3f-4e16-a286-ed7ad3e906b2

3. I've encoded `profile` into a JSON object because it's a stringified json then used the `user` variable to assign it to the `displayName` field.

https://github.com/nhost/hasura-auth/assets/24382050/8d5e281a-495a-4c40-8a1b-d127da0d015a

Please do final tests. I'm not too familiar with the whole project, I might break some stuff. 

Thank you

totzk9:
The [7cc7422](https://github.com/nhost/hasura-auth/pull/425/commits/7cc74226f851acd1a8b836a970d22ce54fc3962c) commit fixes an issue where if an existing Apple user reauthenticates.

Here are the following behaviours:
1. **If new user signs up**, then the `profile` param is a stringified Object thus we need to add a parse it using `JSON.parse(profile)`
2. **If user is already exists and reauthenticates**, then the `profile` param is not a stringified Object causing the `JSON.parse(profile)` to throw an exception.

Solution: I wrapped it in a try catch. Let me know if you want to have requested changes for this one.

Test Video:

https://github.com/nhost/hasura-auth/assets/24382050/a4084ee0-61e9-462d-bcd2-2143dbe7c128

onehassan:
@totzk9 Thank you for the PR 🙌 - Will look into this now

onehassan:
Thank you again @totzk9 for the PR 👍 . I've made some adjustments and added a changeset.
We will merge very soon!

src/routes/oauth/config.ts
```diff
@@ -54,12 +55,22 @@ export const PROVIDERS_CONFIG: Record<
...
+
+      if (profile) {
+        try {
+          const userProfile = JSON.parse(profile);
+
+          displayName = userProfile.name
+            ? `${userProfile.name.firstName} ${userProfile.name.lastName}`
+            : displayName;
+        } catch (error) {
+          logger.warn(error);
```
totzk9, line 68: isn't it unnecessary to always log a warning since existing apple users will always return an empty `profile` based on apple docs?

dbarrosop: I think it's good we log things like this as people may not be aware of this but we should probably add context, logging the error without saying what we were doing might not be very helpful, i.e.:

```
logger.warn(`problem trying to parse user data from apple's reponse: ${error}`);
```

totzk9: Yeah, that makes sense

onehassan: Yeah that's true, will adjust

src/routes/oauth/config.ts
```diff
@@ -51,12 +52,28 @@ export const PROVIDERS_CONFIG: Record<
...
+      if (profile) {
+        try {
+          const userProfile = JSON.parse(profile);
+
+          displayName = userProfile.name
+            ? `${userProfile.name.firstName} ${userProfile.name.lastName}`
+            : displayName;
+        } catch (error) {
+          logger.warn(
+            `Problem trying to parse user data from apple's response: ${error}`
```
dbarrosop, line 69: nitpicking but I'd add `, default to email`

dbarrosop (approved):

src/routes/oauth/config.ts
totzk9: [5b17b5a](https://github.com/nhost/hasura-auth/pull/425/commits/5b17b5ae65daa383755f2a5de718f10578a0699c)

Issue #426, opened 2023-11-02
cannot authenticate with google when using multiple domains


dbarrosop:
Users with multiple domains are having issues when using oauth providers. The issue is as follows:

1. User may have domainA and domainB pointing to hasura-auth
2. domainA is configure as AUTH_SERVER_URL
3. callback url is configured to point to domainA in the aauth provider
4. both domainA and domainB are allowed on the oauth provider

Now, if a user attempts to initiate a login with the oath provider using domainA everything works fine, however, if it uses domainB the last step, when auth tries to exchange the authorization code for an access token, the exchange fails.

This behavior is confirmed at least with Google (using openid scope) and probably Apple.

This is problematic as some apps may need to support both domains for a transitional period of time while their users upgrade their mobile apps.

A potential solution would be to look at the host header and construct the callback url dynamically from it instead of relying on the AUTH_SERVER_URL.

dbarrosop:
Fixed

PR #427, opened 2023-11-03
feat: set oauth origin dynamically based on host header

dbarrosop:
Solves https://github.com/nhost/hasura-auth/issues/426

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -1,3 +1,5 @@
+import { sendError } from '@/errors';
+import { ReasonPhrases } from 'http-status-codes';
 import { json } from 'body-parser';
 import cors from 'cors';
 import express from 'express';
@@ -44,6 +46,18 @@ process.on('uncaughtException', (err, origin) => {
     origin,
   });
   process.exit(1);
+});
+
+/**
+ * GET /healthz
+ * @summary Check if the server is up and running
+ * @return 200 - Success - application/json
+ * @tags General
+ */
+app.get('/healthz', (_req, res) => res.json(ReasonPhrases.OK));
+
+app.use('/', (_req, res) => {
+  return sendError(res,  'bad-request');
 });
 
 export { app };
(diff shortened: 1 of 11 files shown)


src/routes/oauth/utils.ts
```diff
@@ -138,8 +138,7 @@ export const createGrantConfig = (): GrantConfig =>
     },
     {
       defaults: {
-        origin: ENV.AUTH_SERVER_URL,
-        prefix: OAUTH_ROUTE,
+        prefix: `${ENV.AUTH_API_PREFIX}${OAUTH_ROUTE}`,
```
onehassan, line 141: How come we need both the the `AUTH_API_PREFIX` and `OAUTH_ROUTE` here?

src/routes/oauth/index.ts
```diff
@@ -147,6 +147,13 @@ export const oauthProviders = Router()
    * Grant middleware: handle the oauth flow until the callback
    * @see {@link file://./config/index.ts}
    */
+  .use((req, res, next) => {
+    res.locals.grant = {dynamic: {
+        origin: `${req.protocol}://${req.headers.host}`},
+    };
+    next();
+
+  })
```
onehassan, line 156: nitpick: empty line

src/app.ts
```diff
@@ -46,4 +47,8 @@ process.on('uncaughtException', (err, origin) => {
   process.exit(1);
 });
 
+app.use('/', (rwq, res) => {
+  return sendError(res,  'bad-request');
```
onehassan, line 51:
```suggestion
app.use('/', (req, res) => {
  return sendError(res,  'bad-request');
```

src/routes/oauth/utils.ts
dbarrosop: You need to pass to grant the entire prefix where routes will be mounted. If we pass  `OAUTH_ROUTE` we are only passing part of the route.

onehassan (approved):
LGTM 🙌

onehassan:
Can you add a changeset as well

PR #432, opened 2023-11-06
fix(ci): update GitHub workflow to push to Docker Hub with git tag

onehassan:
### Checklist

- [X] No breaking changes
- [X] Tests pass


nunopato (approved):

dbarrosop (approved):

dbarrosop:
will this also push to ECR?

onehassan:
> will this also push to ECR?

Yes it should be, the steps are there. I think it does but not using the correct version number. We will see upon the next release.

PR #440, opened 2023-11-13
fix: mount healthz endpoint before router to avoid overlapping issues

dbarrosop:


onehassan (approved):
LGTM! Can we include a changeset?

PR #453, opened 2023-12-04
fix: added an index on refresh_tokens to improve lookup performance

quentin-decre:
Just adding an index on auth.refresh_tokens (refresh_token_hash, expires_at, user_id)

BEFORE :
image

AFTER :
image

### DIFF
--- a/migrations/00016_index_on_refresh_tokens.sql
+++ b/migrations/00016_index_on_refresh_tokens.sql
@@ -1,4 +1,2 @@
-SET ROLE postgres;
-
 CREATE INDEX IF NOT EXISTS refresh_tokens_refresh_token_hash_expires_at_user_id_idx ON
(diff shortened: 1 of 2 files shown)


dbarrosop:
Thx! Could you add a changeset with `pnpm changeset` or give me access to your fork so I can add it if you prefer?

migrations/00016_index_on_refresh_tokens.sql
```diff
@@ -0,0 +1,4 @@
+SET ROLE postgres;
```
dbarrosop, line 1: this isn't needed and may actually break under some setups

quentin-decre: updated

dbarrosop (approved):

PR #458, opened 2024-01-11
fix: parse metadata during the oauth callback

### LINKED ISSUE #357: Oauth Provider SignIn Metadata not parsed as JSON
While passing the metadata attribute in the query param for the Provider SignIn usecase, the metadata is stored as it is in the metadata column.
No documentation is available for the structure in which metadata is to be passed.
Tried with:
     /provider/{type}?metadata={"key": "value"} 
this got saved in the DB as "{\"key\":\"value\"}"

dbarrosop:
Fixes #357

### DIFF
--- a/.changeset/seven-scissors-poke.md
+++ b/.changeset/seven-scissors-poke.md
@@ -1,0 +1,5 @@
+---
+'hasura-auth': patch
+---
+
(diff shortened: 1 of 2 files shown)


onehassan (approved):
LGTM

PR #462, opened 2024-02-03
feat: added AUTH_REQUIRE_ELEVATED_CLAIM to require elevated permissions for certain actions

dbarrosop:
Adds an option to elevated permissions to perform certain actions: create PATs, change email and/or password, enable/disable MFA and add security keys.

If the option is set to `recommended` the claim check is only performed if the user has a security key attached. If it is set to `required` the only action that won't require the claim is setting a security key for the first time.


nunopato (approved):

src/routes/pat/pat.ts
```diff
@@ -16,10 +16,6 @@ export const createPATHandler: RequestHandler<
   {},
   { metadata: object; expiresAt: Date }
 > = async (req, res) => {
-  if (!req.auth) {
-    return sendError(res, 'unauthenticated-user');
-  }
-
```
onehassan, line 22: Why did we remove this check? I think it's not the same as 'user-not-found'

onehassan (commented):
minor comments

src/routes/pat/pat.ts
dbarrosop: I added the middleware here; which should cover this + the elevation if needed:

https://github.com/nhost/hasura-auth/pull/462/files#diff-0cab04d0c3bdd1b5644679487264bae0931d9a2113af26dc7236bf3977da7137R18

src/routes/user/index.ts
```diff
@@ -67,6 +71,7 @@ router.post(
 router.post(
   '/user/password',
   bodyValidator(userPasswordSchema),
+  // authenticationGate(true, false, (req) => req.body.ticket !== undefined), // this is done in the handler because the handler has an auhtenticated and unauthenticated mode.............
```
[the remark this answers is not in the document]
dbarrosop: actually, I think I want it there because I spent 15 minutes trying to figure why we weren't using the middleware there (see the comment added in the end of the line)

onehassan (approved):

PR #492, opened 2024-03-25
feat: go: migrate /token

nunopato:


go/api/openapi.yaml
```diff
@@ -57,6 +57,30 @@ paths:
...
 
+  /token:
+    post:
+      summary: Refresh the JWT access token
+      tags:
+        - token
+        - access-token
+        - jwt
+      security:
+        - BearerAuthElevated: []
```
dbarrosop, line 68: this doesn't need the elevated security scheme, just the `BearerAuth`

go/controller/workflows.go
```diff
@@ -304,6 +328,79 @@ func (wf *Workflows) GetUserByRefreshTokenHash(
...
+		if err := json.Unmarshal(user.Metadata, &metadata); err != nil {
+			return nil, fmt.Errorf("error unmarshalling user metadata: %w", err)
+		}
+	}
+
+	if _, err := wf.db.UpdateUserLastSeen(ctx, user.ID); err != nil {
+		return nil, fmt.Errorf("error updating user last seen: %w", err)
+	}
+
+	rt, err := wf.GetRefreshTokenByRefreshTokenHash(
```
dbarrosop, line 364: this is redundant, you already did [here](https://github.com/nhost/hasura-auth/pull/492/files#diff-ff137b2c4c6da732fb67bf2e91090217e46acc4be62ffd7a93ddfbd0fb3164dbR16)

In any case, I think there are too many roundtrips here, I think it might be better to have something like:


```sql
-- name: RefreshRefreshToken :one
WITH refresh_token AS (
    UPDATE auth.refresh_tokens
    SET expires_at = $2
    WHERE refresh_token_hash = $1
    RETURNING id, user_id
)
UPDATE auth.users
    SET last_seen = now()
    WHERE id = refresh_token.user_id
    RETURNING refresh_token.id;
```

This single query should:

1. return `pgx.ErrorNoRows` if the hash is wrong so there would be no need to prior check
2. Update the refresh_token expiration
3. Update user's last_seen

this would reduce the endpoint to require only 2 round trips to the database. You may even try to return the roles and do it in a single round-trip but that's probably fine.

go/api/openapi.yaml
nunopato: true, copy/paste issue

go/controller/workflows.go
```diff
@@ -304,6 +308,59 @@ func (wf *Workflows) GetUserByRefreshTokenHash(
...
+	if errors.Is(err, pgx.ErrNoRows) {
+		logger.Warn("invalid refresh token")
+		return &api.Session{}, ErrInvalidRefreshToken //nolint:exhaustruct
+	}
+
+	accessToken, expiresIn, err := wf.jwtGetter.GetToken(
+		ctx, user.ID, roles, user.DefaultRole, logger,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error getting jwt: %w", err)
```
dbarrosop, line 332: log the error and return an `ErrInternalServerError` instead.

go/controller/workflows.go
```diff
@@ -304,6 +308,59 @@ func (wf *Workflows) GetUserByRefreshTokenHash(
...
+	}
+
+	var metadata map[string]any
+	if len(user.Metadata) > 0 {
+		if err := json.Unmarshal(user.Metadata, &metadata); err != nil {
+			return nil, fmt.Errorf("error unmarshalling user metadata: %w", err)
```
dbarrosop, line 338: log the error and return an `ErrInternalServerError` instead.

go/controller/post_token_test.go
```diff
@@ -0,0 +1,168 @@
...
+				Signature: []byte{},
+				Valid:     true,
+			},
+			customClaimer: nil,
+			emailer:       nil,
+			hibp:          nil,
+			jwtTokenFn:    nil,
+		},
+		{
+			name: "invalid refresh token",
```
dbarrosop, line 107: I'd also suggest testing cases like:

1. user is anonymous
2. user is disabled


Mostly because you are reusing `GetUserByRefreshTokenHash` which checks for those and may fail in case you don't want to fail (i.e. it should work for anonymous users but fail for disabled).

go/controller/post_token_test.go
```diff
@@ -0,0 +1,238 @@
...
+				return mock
+			},
+			customClaimer: nil,
+			request: api.PostTokenRequestObject{
+				Body: &api.RefreshTokenRequest{
+					RefreshToken: token.String(),
+				},
+			},
+			expectedResponse: controller.ErrorResponse{
+				Error:   "forbidden-anonymous",
```
dbarrosop, line 203: do we want this behaviour? I haven't checked what we used to do in the nodejs version but I believe an anonymous user should be able to refresh the token

go/controller/workflows.go
```diff
@@ -290,7 +290,11 @@ func (wf *Workflows) GetUserByRefreshTokenHash(
 	)
 	if errors.Is(err, pgx.ErrNoRows) {
 		logger.Error("could not find user by refresh token")
-		return sql.AuthUser{}, ErrInvalidPat //nolint:exhaustruct
+		e := ErrInvalidRefreshToken
+		if refreshTokenType == sql.RefreshTokenTypePAT {
+			e = ErrInvalidPat
+		}
+		return sql.AuthUser{}, e //nolint:exhaustruct
```
dbarrosop, line 297: nitpicking but this alternative is shorter and easier to read as it doesn't need to reference any variable (it also avoids an allocation)

```suggestion
		if refreshTokenType == sql.RefreshTokenTypePAT {
			return sql.AuthUser{}, ErrInvalidPat  //nolint:exhaustruct
		}
		return sql.AuthUser{}, ErrInvalidRefreshToken //nolint:exhaustruct
```

dbarrosop (approved):

PR #513, opened 2024-05-01
feat: allow deep links in redirectTo validation

onehassan:


dbarrosop:
Please, can you add tests for the new schemas supported? Also, same for the go bit (file `validation_test.go`).

src/validation/fields.ts
```diff
@@ -95,6 +95,32 @@ export const redirectTo = Joi.string()
       return value;
     }
 
+    // * Check if the redirectTo is a deep link: start with any sequence of characters
+    // * followed by "://", and then followed by any sequence of characters
+    // * we skip values that start with http: or https://
+    if (/^(?!https?:\/\/).+:\/\/.*/.test(value)) {
```
dbarrosop, line 101: what if the user configures:

```
myapp://redirect
```

but the redirecTo is:

```
myapp://somethingelse
```

?

onehassan: It will be rejected. Good catch 👍 , will adjust

src/validation/fields.ts
```diff
@@ -87,63 +87,57 @@ export const redirectTo = Joi.string()
...
+
+    for (let url of [
+      ...ENV.AUTH_ACCESS_CONTROL_ALLOWED_REDIRECT_URLS,
+      ENV.AUTH_CLIENT_URL,
+    ]) {
+      switch (true) {
+        case url.endsWith('/**'):
+          break;
+        case url.endsWith('/*'):
+          url += '*';
```
dbarrosop, line 103: will this modify the original string?

dbarrosop (approved):

src/validation/fields.ts
onehassan: this will mutate the `url` but here we're just looping through a copy of `AUTH_ACCESS_CONTROL_ALLOWED_REDIRECT_URLS` and so the array of the allowed redirect urls won't be mutated

PR #525, opened 2024-06-11
/token failing for user with Empty email

blib:
/token is failing email validation when email is set to null


dbarrosop:
what is the use case?

blib:
Anonymous user with null email trying to refresh token fails on email validation (which is null)

go/api/openapi.yaml
```diff
@@ -519,7 +519,6 @@ components:
         email:
           description: A valid email
           example: john.smith@nhost.io
-          format: email
```
dbarrosop, line 522: please, don't remove this check as this still has to be a valid email, instead remove `email` from "required" below so this is optional and we can simply skip it in the response payload.

dbarrosop:
could you rebase also, please? I have a fix in main that I think/hope will fix the CI

etodanik:
@dbarrosop what are the chances for it to land? It became a breaking issue for us, we can't use nhost unless this lands because our app is phone-only authentication.

go/api/openapi.yaml
blib: The proposed change will not achieve the desired outcome and would necessitate a much broader modification. Removing the requirement would transform the email field into a pointer in the code generation process, resulting in a massive overhaul. While there is already an email validator function `ValidateEmail` in `controller.go`, which can be extended with regular expressions if needed, the current implementation appears sufficient.

Additionally, incorporating validation during marshaling is generally considered an anti-pattern, as it leads to 500 errors without providing meaningful feedback to the user.

blib: I'm not that familiar with oapi-codegen but on first sight it looks like there is a nullable type wrapper that could be used for proposed semantics 

`nullable.Nullable[types.Email]`

dbarrosop:
Thanks for raising the issue and for the PR. This should be fixed in #529

go/api/openapi.yaml
dbarrosop: if you take a look at #529 you will see the change isn't big at all (probably the same size after you had fixed the tests due to the type change) but it is also simpler. It is also more correct as the format is not only for the sake of generating the API, it is also useful for tooling/generating clients so informing them that something isn't required but if present should be an email is more useful than removing the restriction.

Thanks again for raising the issue and coming up with an initial solution, it certainly made things simpler even if the proposed solution didn't make it for one reason or another.

blib:
thank you for you great work! Your fix is much cleaner!

PR #617, opened 2025-03-14
feat: refresh session with new refresh token

### LINKED ISSUE #416: change refresh-token behavior to expire after one use

antoinecaputo:
This PR implements token rotation for refresh tokens, improving security by invalidating the old refresh tokens when a new one is generated.

- Edit the `RefreshTokenAndGetUserRoles` query to update both the `expires_at` and `refresh_token_hash` fields
- No additional database calls

### Checklist

- [ ] No breaking changes
- [x] Tests pass

### Breaking changes

It may affect clients that don't store the new refresh token from the `/token` session response.

Clients must store the new refresh token received in each session response.

If you are using nhost's official SDK this change is transparent and has no impact.

#### Coverage

The `invalid refresh token` test cover the case where a deleted refresh token is used.

#### Missing uuid mock

The `uuid.New()` call in `UpdateSession` workflow isn't mocked in tests.

Therefore, the `RefreshTokenAndGetUserRolesParams` matcher will only expect the `NewRefreshTokenHash` value to have changed.

However, the `RefreshTokenId` field is correctly updated with the mock return.

### Documentation

Updated `docs/workflows/refresh-token.md` to add the refresh tokens update.

Closes #416


go/controller/workflows.go
```diff
@@ -387,38 +387,44 @@ func pgtypeTextToOAPIEmail(pgemail pgtype.Text) *types.Email {
...
 }
 
-func (wf *Workflows) UpdateSession( //nolint:funlen
+func (wf *Workflows) RefreshSession( //nolint:funlen
 	ctx context.Context,
 	user sql.AuthUser,
-	refreshToken string,
 	logger *slog.Logger,
 ) (*api.Session, *APIError) {
-	userRoles, err := wf.db.RefreshTokenAndGetUserRoles(ctx, sql.RefreshTokenAndGetUserRolesParams{
```
dbarrosop, line 396: I actually think this could be much easier and efficient. I believe we could keep the RefreshTokenAndGetUserRoles method but pass a new newly generated token:

```go
func (wf *Workflows) UpdateSession(
	ctx context.Context,
	user sql.AuthUser,
	logger *slog.Logger,
) (*api.Session, *APIError) {
    refreshToken := uuid.New()
	userRoles, err := wf.db.RefreshTokenAndGetUserRoles(ctx, sql.RefreshTokenAndGetUserRolesParams{
		RefreshTokenHash: sql.Text(hashRefreshToken([]byte(refreshToken))),
		ExpiresAt: sql.TimestampTz(
			time.Now().Add(time.Duration(wf.config.RefreshTokenExpiresIn) * time.Second),
		),
	})
    ...
```

I think rest would be the same. The main issue is that your changes make this endpoint slightly less performant due to the added call to the database (we went from 2 to 4 so it adds two more RTTs)

dbarrosop:
Good job, glad you managed to figure out how to contribute, documentation here sucks. I added a comment as I think we can simplify the change and keep the performance of this endpoint.

go/sql/query.sql
```diff
@@ -249,22 +249,6 @@ INSERT INTO auth.refresh_tokens (user_id, refresh_token_hash, expires_at, type,
 VALUES ($1, $2, $3, $4, $5)
 RETURNING id;
 
--- name: RefreshTokenAndGetUserRoles :many
```
dbarrosop, line 252: Instead of removing this one you could do this:

```sql
-- name: RefreshTokenAndGetUserRoles :many
WITH refreshed_token AS (
    UPDATE auth.refresh_tokens
    SET expires_at = $2, refresh_token_hash = sqlc.arg(newRefreshToken)
    WHERE refresh_token_hash = sqlc.arg(oldRefreshToken)
    RETURNING id AS refresh_token_id, user_id
),
updated_user AS (
    UPDATE auth.users
    SET last_seen = now()
    FROM refreshed_token
    WHERE auth.users.id = refreshed_token.user_id
)
SELECT refreshed_token.refresh_token_id, role FROM auth.user_roles
RIGHT JOIN refreshed_token ON auth.user_roles.user_id = refreshed_token.user_id;
```

antoinecaputo:
Thanks for the review, it makes sense to reduce the number of operations.
I will take a look at it.

antoinecaputo:
Change applied @dbarrosop 
I have updated the PR description to match the new behavior

dbarrosop:
amazing, this looks perfect now. I will build an image and test it soon

dbarrosop:
Ok, I pushed a minor fix and deployed the image to:

https://react-apollo.example.nhost.io

You can see it working by inspecting the localstorage and reloading the page. Will do some more testing during the week before merging.

dbarrosop (approved):

dbarrosop:
Thanks again for this contribution

PR #631, opened 2025-05-10
fix: idtoken panic error

### LINKED ISSUE #628: panic error on id token signup
Hi, I was working on a fix for an issue with the dart sdk during the id token sign up (https://github.com/nhost/nhost-dart/pull/162/commits/12850e90751a818a8cf441b296e123e4be3581ff). however, i'm now encountering a panic error log in hasura auth (v0.38.0) on sign up (sign in works fine):

Image

I'm using the dart sdk like so:
```
await nhost.auth.signInIdToken(
      provider: "google",
      idToken: idToken,
      locale: "en",
      displayName: "-",
    );
```

PrinzKenny1:
Fixes #628 

Before submitting this PR:

### Checklist

- [x] No breaking changes
- [x] Tests pass

### Breaking changes

### Tests

### Documentation

### DIFF
--- a/go/controller/post_signin_idtoken.go
+++ b/go/controller/post_signin_idtoken.go
@@ -126,6 +126,7 @@ func (ctrl *Controller) postSigninIdtokenSignup( //nolint:ireturn
 	session, apiErr := ctrl.wf.SignupUserWithFn(
 		ctx,
 		profile.Email,
+		profile.EmailVerified,
 		req.Body.Options,
 		false,
 		ctrl.postSigninIdtokenSignupWithSession(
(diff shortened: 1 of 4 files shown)


dbarrosop:
Thanks for the PR, interestingly we had already addressed this (or a similar issue at least) on a different PR that is currently on the works to migrate the social login to go:

https://github.com/nhost/hasura-auth/pull/625

If you don't mind let's wait a bit until the other one is fixed and we can see if that one addresses the issue. Thanks again for the contribution.

I am curious though, do you know how is it possible that an idtoken with google/apple leads to an unverified email? I didn't expect that to happen with these two providers.

PrinzKenny1:
@dbarrosop i just looked thru ur PR and indeed it should fix this issue as well. I also added a comment there bc I guess it's missing the brackets 🤔 
The emails themselves should I guess nearly always be verified, i found that post abt it https://jpassing.com/2021/01/27/what-does-the-email_verified-claim-indicate-in-google-idtokens/. Also:
grafik
grafik
On the bottom it says: "You'll have to confirm that this email belongs to you"

Tho I doubt it's possible to use the google account without verifying that first, but i haven't tested it

dbarrosop:
> I also added a comment there bc I guess it's missing the brackets
they are not needed in golang and && takes precedence over || mathematically so they are redundant in this case but they certainly increase readability so they are probably worth adding

> On the bottom it says: "You'll have to confirm that this email belongs to you"

interesting, anyway, it should be handled regardless as while it might or might not be possible with google/apple it might with others, I was just curious. Thanks again.

